PR #12, opened 2026-01-09
[fix] fix grafana start cmd

wz1qqx:
fix typo error

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -124,7 +124,7 @@ The `examples/metric/` directory provides a ready-to-use monitoring stack.
 
 ```bash
 cd examples/metric
-docker compose up -d
+docker-compose up -d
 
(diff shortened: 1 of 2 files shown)


docs/metrics.md
```diff
@@ -124,7 +124,7 @@ The `examples/metric/` directory provides a ready-to-use monitoring stack.
 
 ```bash
 cd examples/metric
-docker compose up -d
+docker-compose up -d
```
xiaguan: Both `docker compose up -d` (Compose V2, integrated as a Docker CLI plugin) and `docker-compose up -d` (standalone Compose V1) are valid. We can keep both commands in the documentation to accommodate users on different Docker setups.

xiaguan (commented):
just a nit

PR #31, opened 2026-01-15
feat(server): simplify server startup command and document all options

xiaguan:
- Add default-run = "pegaflow-server" to allow `cargo run -r` without --bin
- Update README to show simplified command (cargo run -r)
- Document all 13 server CLI options with defaults
- Highlight pool-size as most commonly customized parameter

This improves user experience by reducing command verbosity while maintaining full configurability through optional parameters.


wz1qqx (approved):
LGTM! Github Docs maybe a good first issue? @xiaguan

PR #37, opened 2026-01-16
docs(readme): generalize description and add SGLang integration status

xiaguan:
- Replace vLLM-specific wording with generic "LLM inference engines"
- Move Pegasus tagline under logo
- Add Framework Integration table with SGLang PR link


wz1qqx (approved):
LGTM！

Issue #11, opened 2026-01-08
[perf]: Improve Memory Utilization of Offset Allocator


xiaguan:
In production deployments with TB-scale memory, we observed the offset allocator achieving only ~88% minimum utilization, resulting in hundreds of GB of wasted memory. The root cause is currently unknown.

**Investigation Plan:**

- Write a benchmark based on [allocator_bench.cpp](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/benchmarks/allocator_bench.cpp) to observe memory utilization fluctuations under steady-state LLM KV cache workloads
- Expected baseline utilization before optimization: ~90%

**Workload Characteristics:**

- Fixed page size per model (assuming single model for now)
- Allocation size: 20-100 pages per request (random)
- Example: Qwen 0.6B with 2K context ≈ 224MB

PR #39, opened 2026-01-17
Benchmark for memory utilization of offset allocator (#11)

swaploard:
KV Cache Fragmentation Benchmark (Issue #11)

This PR adds a Rust benchmark for ScaledOffsetAllocator that simulates a steady-state LLM KV-cache workload, closely mirroring allocator_bench.cpp and production usage.

**Workload Model**

Fixed page size per model (128 KB)

Allocation size: 20–100 pages per request (random)

Long-running steady state with periodic churn

Prefill to ~85% utilization, then operate under pressure

Tracks:

Utilization over time

Internal fragmentation

External fragmentation

Allocation failures with free memory present

Allocation latency

This reproduces the production symptom:

allocator reaches ~88–90% minimum utilization despite large amounts of free memory remaining.

### DIFF
--- a/pegaflow-core/benches/allocator_bench.rs
+++ b/pegaflow-core/benches/allocator_bench.rs
@@ -1,6 +1,3 @@
-/* This is a Rust benchmark for ScaledOffsetAllocator that simulates a steady-state LLM KV-cache workload, 
-closely mirroring allocator_bench.cpp and production usage. */
-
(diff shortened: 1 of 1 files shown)


xiaguan:
This isn't what I expected. Perhaps you could check https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/benchmarks/allocator_bench.cpp for reference. Additionally, please add this benchmark to Cargo.

PR #44, opened 2026-01-20
test: add E2E correctness canary test for vLLM integration

xiaguan:
Add a golden test that verifies PegaFlow produces identical outputs to baseline vLLM. This is the core contract test:

- Baseline vLLM (no KV connector) output
- PegaFlow cold cache output
- PegaFlow warm cache output

All three must be bit-exact identical.


wz1qqx (approved):
LGTM！

PR #51, opened 2026-01-23
fix(metrics): fix pinned_for_load leak tracking and improve memory metrics

xiaguan:
## Summary
- Add `inflight_bytes` metric to track memory in inflight blocks (write path)
- Add `pinned_for_load_unique_bytes` to track pinned blocks that prevent eviction
- Fix `cache_lookup_many` to use Entry API and properly track pin removal with `error!` + `debug_assert!` for invariant violations
- Add metrics helper functions for consistent +/- tracking
- Update dashboard: Memory Accounting panel shows `pool_used` vs `cache_resident` gap (caused by allocation fragmentation from batch allocation sharing)
- Clean up redundant dashboard panels

## Background
Discovered that `pool_used_bytes` (~30GB) was significantly higher than `cache_resident_bytes` (~13GB). Root cause: batch allocations are shared across multiple SealedBlocks, and only when ALL blocks sharing an allocation are evicted can the memory be freed.

## Test plan
- [x] Verify metrics are exported correctly via Prometheus
- [x] Check dashboard displays Memory Accounting correctly
- [x] Confirm `pool_used` and `cache_resident` gap is visible

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -44,6 +44,10 @@ PegaFlow exposes the following metrics for monitoring KV cache operations:
   - Total pinned memory pool capacity in bytes
   - Use case: Derive pool utilization
 
+- **pegaflow_pool_largest_free_bytes** (Gauge)
+  - Largest contiguous free region in pinned pool (fragmentation signal)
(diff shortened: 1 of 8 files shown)


wz1qqx (approved):
LGTM!

PR #78, opened 2026-01-29
feat(numa): NUMA-aware pinned memory allocation for optimal GPU affinity

xiaguan:
## Summary

Add NUMA-aware pinned memory allocation to optimize GPU-CPU transfer bandwidth on multi-socket systems.

### Key Changes

- **Per-NUMA pinned pools**: `NumaAwarePinnedPools` allocates memory on each NUMA node using first-touch policy (thread pinned to target node during allocation)
- **GPU NUMA affinity tracking**: `GpuContext` tracks each GPU's preferred NUMA node via `nvidia-smi topo --get-numa-id-of-nearby-cpu`
- **Worker thread pinning**: Load/save worker threads are pinned to GPU's local NUMA node for optimal memory locality during D2H/H2D transfers
- **Unified PinnedAllocator**: Supports both global (single pool) and NUMA-aware (per-node pools) modes
- **Topology consolidation**: Merged `numa/topology.rs` into `numa.rs`

### API Improvements

- `StorageConfig`: `numa_nodes: Option<Vec<NumaNode>>` → `enable_numa_affinity: bool` (clearer semantics)
- CLI: `--disable-numa-affinity` → `--numa-affinity` (default: `true`, use `--numa-affinity=false` to disable)

### Why This Matters

On multi-socket systems (e.g., dual-socket servers with multiple GPUs), memory access latency varies significantly based on which NUMA node the memory is allocated on. By allocating pinned memory on the same NUMA node as the GPU's PCIe controller, we minimize cross-socket memory traffic and achieve optimal transfer bandwidth.

## Test plan

- [ ] Verify build passes (`cargo check`, `cargo test`)
- [ ] Test on single-NUMA system (should use global pool)
- [ ] Test on multi-NUMA system (should create per-node pools)
- [ ] Verify `--numa-affinity=false` disables NUMA-aware allocation

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pegaflow-core/src/pinned_pool.rs
+++ b/pegaflow-core/src/pinned_pool.rs
@@ -298,7 +298,11 @@ impl NumaAwarePinnedPools {
     /// # Behavior
     /// - Each NUMA node gets `total_capacity / num_nodes` bytes
     /// - Pools are allocated on threads pinned to their respective NUMA nodes
-    /// - If a NUMA pool allocation fails, that node is skipped (logged as warning)
+    ///
+    /// # Panics
+    /// Panics if any NUMA pool allocation fails. This is a fail-fast behavior
+    /// to prevent silent partial failures that would cause mysterious allocation
+    /// errors later when GPUs try to use the missing pool.
     pub fn new(
         total_capacity: usize,
         numa_nodes: &[NumaNode],
@@ -346,7 +350,12 @@ impl NumaAwarePinnedPools {
                     pools.insert(node_id, Arc::new(pool));
                 }
                 Err(e) => {
-                    warn!("Failed to create pool on NUMA{}: {}", node_id, e);
+                    panic!(
+                        "Failed to create pool on NUMA{}: {}. \
+                         This is a fatal error during initialization. \
+                         Please check system resources or disable NUMA affinity.",
+                        node_id, e
+                    );
                 }
             }
         }


pegaflow-core/src/storage.rs
```diff
@@ -399,40 +428,70 @@ impl StorageEngine {
...
         loop {
-            if let Some(allocation) = self.pinned_pool.allocate(size) {
-                return Some(Arc::new(allocation));
+            // Try to allocate from the unified allocator
+            if let Some(alloc) = self.allocator.allocate(size, node) {
+                return Some(alloc);
             }
 
+            // Allocation failed, try to reclaim memory
             let (freed_blocks, freed_bytes, largest_free) =
```
jimmy-evo: The eviction logic here is problematic for the NUMA scenario.

`reclaim_until_allocator_can_allocate` evicts blocks globally (from `LRU`/`TinyLFU`), but `allocate` needs space on a *specific* NUMA node.

If I am trying to allocate on `Node 0`, but the cache eviction victim happens to be on `Node 1`, I free up space on Node 1, but Node 0 remains full. The allocation loop will then retry, see `largest_free` on Node 0 is still insufficient, call reclaim again... and potentially evict the entire cache on other nodes without ever freeing space on the target node.

**Suggestion:**
The reclamation logic needs to be NUMA-aware.
1. `reclaim_until_allocator_can_allocate` should accept a `target_numa_node`.
2. The eviction policy needs to support "evict from specific pool" or the allocator needs to be able to tell if a freed block actually helped.

Without this, you risk "thrashing" the cache on the wrong nodes while the target node stays full.

wz1qqx (approved):
LGTM！

PR #89, opened 2026-02-02
refactor(ssd): unify ring buffer state with two-phase commit

xiaguan:
## Summary

Refactors the SSD cache layer to eliminate race conditions between asynchronous writes and ring buffer eviction by introducing a unified `SsdRingBuffer` with two-phase commit semantics.

## Motivation

Previously, the SSD cache had split state between `StorageInner.ssd_index` and `SsdState.head`, with the writer thread directly managing the ring buffer. This could lead to race conditions where:
1. A block is being written to SSD (IO in flight)
2. Ring buffer wraps around and evicts the same logical offset
3. Prefetch reads the location and gets stale/corrupt data

### Core Architecture
- **New `SsdRingBuffer` struct**: Combines head/tail pointers, FIFO ordering (`VecDeque`), and fast lookup (`HashMap`) into one coherent structure
- **Two-phase commit**: New blocks start in `Writing` state, transition to `Committed` only after successful IO. Reads only see `Committed` entries with valid offsets.
- **Unified callbacks**: `SsdStorageHandle` now exposes `prepare()` (allocate + insert Writing) and `commit()` (finalize or abort) instead of direct tail/head manipulation

### API Cleanup
- Removed unused `Read`/`Write` IoType variants from `uring.rs` (only `Readv`/`Writev` are used)
- Made SSD-internal types `pub(crate)` instead of `pub`

### Testing
- Added 10+ unit tests covering:
  - Contiguous allocation with wrap-around
  - Tail advancement and entry pruning
  - Two-phase commit (success/failure/expired cases)
  - Duplicate key filtering
  - Ghost entry cleanup

## Backward Compatibility

No breaking changes to public APIs. This is an internal refactoring of the SSD cache subsystem.

## Checklist

- [x] Unit tests added for new components
- [x] Existing functionality preserved (no behavioral changes to cache hit/miss semantics)
- [x] Code follows project style guidelines

### DIFF
--- a/pegaflow-core/src/ssd_cache.rs
+++ b/pegaflow-core/src/ssd_cache.rs
@@ -1,6 +1,6 @@
 use futures::stream::{FuturesOrdered, FuturesUnordered, StreamExt};
 use log::{debug, warn};
-use std::collections::HashSet;
+use std::collections::{HashMap, VecDeque};
 use std::path::PathBuf;
 use std::sync::{Arc, Weak};
 use std::time::Instant;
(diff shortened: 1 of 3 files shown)


wz1qqx (approved):
LGTM！

PR #103, opened 2026-02-09
fix: allow configurable grpc memssage size 

Kevin-XiongC:


Kevin-XiongC:
> We could hard code it to 64MB using a const in the code—no need for an env variable, keeping complexity low. Otherwise, LGTM.

done

PR #106, opened 2026-02-11
perf(save): optimize save path with batched GPU copies and hash-first insert

xiaguan:
## Summary

- **Batch GPU copies**: Restructured save from per-layer sequential processing to 4-phase pipeline. All 184 layers' D2H copies issued on a single CUDA stream with one synchronization, reducing CUDA sync points from 184 to 1.
- **Hash-first insert**: All layers share the same block hashes (token-content-based). Filter and insert now operate on unique hashes (1,563) instead of per-layer (184 × 1,563 = 287,592), reducing HashMap ops and heap allocations by 184x.
- **Log cleanup**: Downgraded profiling `warn!` logs to `debug!`/`info!` for production readiness.

### Performance (100k tokens, page_size=64, tp=8)

| Metric | Before | After | Speedup |
|--------|--------|-------|---------|
| Engine time (best TP) | 1,153ms | 200ms | 5.8x |
| Engine time (worst TP) | 1,168ms | 246ms | 4.7x |
| RPC e2e | ~1,250ms | 282-338ms | ~4x |

Phase 3 (GPU copy) is now at PCIe hardware floor (~134ms for 4.7GB).

## Test plan

- [x] `cargo check` passes
- [x] `cargo fmt --check` passes
- [ ] Benchmark with 100k input tokens, page_size=64, tp=8

### DIFF
--- a/pegaflow-core/src/offload.rs
+++ b/pegaflow-core/src/offload.rs
@@ -99,9 +99,7 @@ pub(crate) struct RawSaveBatch {
 ///
 /// Returns `(entries, total_bytes, total_blocks)` where entries is grouped
 /// by hash: `Vec<(BlockKey, Vec<(slot_id, Arc<LayerBlock>)>)>`.
-pub(crate) fn build_insert_entries(
-    batch: &RawSaveBatch,
-) -> (InsertEntries, u64, usize) {
+pub(crate) fn build_insert_entries(batch: &RawSaveBatch) -> (InsertEntries, u64, usize) {
     use std::collections::HashMap;
 
     let mut hash_entries: HashMap<Vec<u8>, Vec<(usize, Arc<LayerBlock>)>> = HashMap::new();


wz1qqx (approved):
LGTM！

pegaflow-core/src/block.rs
```diff
@@ -247,46 +223,26 @@ impl InflightBlock {
...
-                got: total_slots,
-            });
+    /// Insert a slot idempotently. Duplicate inserts are no-ops.
+    ///
+    /// Returns `true` if all slots are now filled (ready to seal).
+    pub fn insert_slot(&mut self, slot_id: usize, block: Arc<LayerBlock>) -> bool {
+        debug_assert!(
+            slot_id < self.total_slots,
+            "slot_id {} must be < total_slots {}",
+            slot_id,
```
jimmy-evo: 🔴 **Critical: `debug_assert!` is a no-op in release builds — silent UB risk**

`debug_assert!` compiles away in release mode. If `slot_id >= self.total_slots`, the subsequent `self.slots[slot_id]` access on line 234 will **panic with an unhelpful index-out-of-bounds message** (or worse, if `slots` was allocated with excess capacity, it could silently corrupt state).

The old code returned `Err(BlockInsertError::SlotOutOfBounds)` which allowed graceful error handling.

**Recommendation**: Either use a full `assert!` for a clear panic message in release, or add a guard:
```rust
if slot_id >= self.total_slots {
    warn!("insert_slot: slot_id {} out of bounds (total={})", slot_id, self.total_slots);
    return false;
}
```
Given this is on the insert-worker path (not the ultra-hot GPU copy path), the branch cost is negligible.

pegaflow-core/src/storage.rs
```diff
@@ -1099,3 +1036,223 @@ impl StorageEngine {
...
+        for (slot_id, block) in slots {
+            let footprint_bytes = block.memory_footprint();
+            completed = inflight_block.insert_slot(slot_id, block);
+            inflight_bytes_added = inflight_bytes_added.saturating_add(footprint_bytes);
+
+            if completed {
+                break;
+            }
+        }
+
```
jimmy-evo: 🟡 **Inflight bytes metric overcounts on duplicate slot inserts**

When `insert_slot` is a no-op (duplicate slot, which is idempotent), `inflight_bytes_added` still accumulates `footprint_bytes`. Over time this will cause the `inflight_bytes` metric to drift upward, making capacity monitoring unreliable.

**Recommendation**: Have `insert_slot` return a `(completed: bool, inserted: bool)` tuple, or check the slot before insertion:
```rust
let was_empty = inflight_block.slots[slot_id].is_none(); // or add a method
completed = inflight_block.insert_slot(slot_id, block);
if was_empty {
    inflight_bytes_added = inflight_bytes_added.saturating_add(footprint_bytes);
}
```

python/src/lib.rs
```diff
@@ -274,14 +274,10 @@ impl EngineRpcClient {
...
 
-        const MAX_GRPC_MESSAGE_SIZE: usize = 64 * 1024 * 1024; // 64 MiB
-
         let channel = rt
             .block_on(endpoint_cfg.connect())
             .map_err(|err| transport_connect_error(&endpoint, err))?;
-        let client = EngineClient::new(channel)
-            .max_decoding_message_size(MAX_GRPC_MESSAGE_SIZE)
-            .max_encoding_message_size(MAX_GRPC_MESSAGE_SIZE);
+        let client = EngineClient::new(channel);
```
jimmy-evo: 🟡 **gRPC message size limits accidentally removed**

The previous `max_decoding_message_size(64 MiB)` and `max_encoding_message_size(64 MiB)` settings have been dropped. Tonic's default limit is **4 MiB**. With 184 layers × ~1,563 blocks × hash data, the `BatchSave` request payload can easily exceed 4 MiB, causing **runtime `Status::resource_exhausted` errors** for large save batches.

This looks accidental — recommend reinstating the limits:
```rust
let client = EngineClient::new(channel)
    .max_decoding_message_size(MAX_GRPC_MESSAGE_SIZE)
    .max_encoding_message_size(MAX_GRPC_MESSAGE_SIZE);
```

jimmy-evo (commented):
## 🔍 Post-Merge Code Review

Great work on this PR — the 4-phase batched pipeline is a clean architectural improvement and the 4.7–5.8× speedup is impressive. Reducing CUDA sync points from 184→1 and deduplicating hash operations across layers are high-impact optimizations.

### Summary of Findings

| Severity | Count | Key Items |
|----------|-------|-----------|
| 🔴 Critical | 3 | Unbounded channel (pinned memory exhaustion risk), insert worker blocking Tokio runtime, `debug_assert!` no-op in release |
| 🟡 Important | 3 | Inflight bytes metric overcount, SSD offload on rejected/duplicate cache inserts, gRPC message size limits removed |
| 🟠 Suggestions | 3 | Heap allocations in hash filter under mutex, redundant `unsafe impl Send`, missing unit tests for `build_insert_entries` |

### Recommended Follow-up Priority

1. **[P0]** Reinstate gRPC message size limits in `python/src/lib.rs` — this looks accidental and will cause runtime failures for large batches
2. **[P0]** Switch unbounded channel to bounded with backpressure — pinned memory is finite
3. **[P1]** Move insert worker to a dedicated OS thread (`std::thread::spawn`) to avoid blocking Tokio
4. **[P1]** Replace `debug_assert!` with a proper guard in `insert_slot`
5. **[P2]** Fix metric overcount and SSD duplicate write issues
6. **[P2]** Add unit tests for `build_insert_entries` grouping logic

### What's Done Well
- ✅ Clean module extraction (`offload.rs`) with well-documented phases
- ✅ Insert worker actor pattern eliminates mutex contention on hot path
- ✅ Per-phase timing in debug logs is excellent for profiling
- ✅ New tests for `contains_key` frequency behavior and cache filter
- ✅ Log level cleanup for production readiness

pegaflow-core/src/storage.rs
xiaguan: Confirmed — this is a real bug. When duplicate hashes exist within a single layer, `build_insert_entries` groups them under the same hash with the same `slot_id`. The second `insert_slot` is a no-op but `inflight_bytes_added` still accumulates, causing a positive drift in the metric over time. Tracked in #111.

pegaflow-core/src/gpu_worker.rs
```diff
@@ -49,6 +55,9 @@ pub struct SaveBlock {
 // by PinnedAllocation. The caller ensures the backing memory stays alive
 // until the task completes.
 unsafe impl Send for SaveBlock {}
+// SAFETY: SaveLayerData contains SaveBlocks which hold raw pointers to pinned
+// memory. Same safety guarantees as SaveBlock apply.
+unsafe impl Send for SaveLayerData {}
```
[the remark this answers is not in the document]
xiaguan: Good catch — will clean this up.

PR #119, opened 2026-03-02
fix(core): per-slot NUMA affinity for SSD prefetch allocation

### LINKED ISSUE #113: SSD prefetch NUMA affinity is per-block but should be per-slot for PP scenarios
## Problem

`SsdIndexEntry` stores a single `numa_node` per block:

```rust
// pegaflow-core/src/ssd_cache.rs
pub(crate) struct SsdIndexEntry {
    pub begin: u64,
    pub len: u64,
    pub file_offset: u64,
    pub slots: Vec<SlotMeta>,
    pub numa_node: NumaNode,  // ← single NUMA node for the entire block
}
```

A `SealedBlock` has `total_slots = num_layers * tp_size` slots. In TP, different ranks are treated as different layers (e.g., `layer.1.tp_rank_0`, `layer.1.tp_rank_1`), and PP is supported — different PP stages run on different GPUs, potentially on different NUMA nodes.

This means a single block's slots can originate from **multiple GPUs across different NUMA nodes**.

### Save → SSD path

When an `InflightBlock` is sealed in `process_insert_batch`, the `numa_node` passed to `send_ssd_batch` comes from whichever `RawSaveBatch` triggered the final seal:

```rust
// storage.rs:1091-1096
let numa_node = batch.numa_node;  // last caller's GPU NUMA node
// ...
engine.send_ssd_batch(numa_node, &sealed_blocks);
```

This records a single (potentially arbitrary) NUMA node for a block whose slots span multiple NUMA nodes.

### SSD prefetch → load path

The prefetch dispatcher groups requests by `SsdIndexEntry.numa_node` and allocates from the corresponding NUMA pool:

```rust
// ssd_cache.rs — ssd_prefetch_dispatcher
// Groups by req.entry.numa_node → allocates from that NUMA pool
handle.allocate(total_size, group.numa_node)
```

The entire block's prefetch buffer is allocated on one NUMA node, but the data will be loaded to GPUs on potentially different NUMA nodes. This defeats the NUMA locality optimization for cross-node slots.

## Impact

- On multi-socket systems with PP, prefetched KV cache blocks may reside on a remote NUMA node relative to the target GPU, causing H2D transfers to traverse QPI/UPI with reduced bandwidth.
- The `numa_node` recorded in `SsdIndexEntry` may be non-deterministic (depends on which PP stage's save completes last), making performance unpredictable.

## Possible approaches

1. **Per-slot NUMA node**: Store `numa_node` per `SlotMeta` instead of per `SsdIndexEntry`. Prefetch allocates each slot from its respective NUMA pool.
2. **Load-target aware allocation**: Instead of using the save-time NUMA node, allocate prefetch buffers based on the **load target GPU's** NUMA node (the GPU that will consume the data).
3. **Accept the limitation**: Document that NUMA-aware SSD prefetch is only optimal for TP-only (no PP) deployments, where all slots in a block come from the same GPU/NUMA node.

xiaguan:
## Summary

Fixes #113

- Track per-slot NUMA affinity through the full pipeline: `InflightBlock` → `SealedBlock` → `SlotMeta` → prefetch dispatcher
- Prefetch dispatcher groups all slots across blocks by NUMA node, allocates one chunk per group, scatters to per-slot offsets
- If any NUMA group allocation fails, fail the entire batch (all blocks share the same slot layout)
- Consolidate split/contiguous `LayerBlock` construction into `SlotMeta::make_layer_block`
- Remove unused serde derives from `SlotMeta`

## Test plan

- [x] `cargo test` — 16 ssd_cache tests pass (including 3 new NUMA grouping tests)
- [x] `cargo clippy` — clean
- [x] `./scripts/check.sh` — pass

### DIFF
--- a/pegaflow-core/src/block.rs
+++ b/pegaflow-core/src/block.rs
@@ -5,6 +5,7 @@
 use std::sync::Arc;
 use std::time::Instant;
 
+use crate::numa::NumaNode;
 use crate::pinned_pool::PinnedAllocation;
 
 // ============================================================================
@@ -154,6 +155,9 @@ unsafe impl Sync for LayerBlock {}
 pub struct SealedBlock {
     slots: Box<[Arc<LayerBlock>]>,
     footprint: u64,
+    /// Per-slot NUMA affinity, carried from InflightBlock for SSD write path.
+    /// Empty when reconstructed from SSD prefetch (NUMA info lives in SlotMeta).
+    slot_numas: Vec<NumaNode>,
 }
 
 impl SealedBlock {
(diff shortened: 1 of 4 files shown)


wz1qqx (approved):
lgtm！

PR #129, opened 2026-03-06
feat(connector): enable cross-layer blocks by default

xiaguan:
## Summary
- Switch `PEGAFLOW_CROSS_LAYER_BLOCKS` default from `"0"` to `"1"`, so cross-layer unified tensor is used when the attention backend supports it (reduces cudaMemcpy calls from N to 1 per save/load)
- Add `cross_layer_blocks` as a namespace factor in `derive_namespace()` to isolate cross-layer and per-layer block data, preventing hash collisions when switching between modes on the same server

## Test plan
- [x] Run `bench_kv_cache.py` with default (cross-layer on) and verify single-layer registration in logs
- [x] Run with `PEGAFLOW_CROSS_LAYER_BLOCKS=0` and verify per-layer registration still works
- [x] Verify namespace hash differs between cross-layer and per-layer modes

### DIFF
--- a/pegaflow-core/src/lib.rs
+++ b/pegaflow-core/src/lib.rs
@@ -15,6 +15,7 @@ pub mod block;
 mod cache;
 pub mod gpu_worker;
 pub mod instance;
+#[allow(dead_code)]
 pub mod internode;
 pub mod logging;
 mod metrics;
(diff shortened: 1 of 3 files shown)


wz1qqx (dismissed):
LGTM！

wz1qqx (approved):
LGTM!

PR #140, opened 2026-03-09
chore: unify rust and python package versions

xiaguan:
## Summary
- unify the workspace Rust version with the Python package release version
- expose the compiled package version through Rust CLIs and 
- add a CI check to keep Rust and Python version declarations in sync

- local version consistency script matching CI logic


wz1qqx (approved):
LGTM！

PR #164, opened 2026-03-20
fix(python): remove TransferEngine and relax ipc_wrapper contiguity check

xiaguan:
## Summary

- Remove `TransferEngine` from `__init__.py`, `pegaflow.pyi`, and delete `sglang/transfer_engine.py` — the Rust binding was already removed but Python references were left behind, causing `import pegaflow` to fail with `ImportError`
- Drop `is_contiguous()` assertion in `CudaIPCWrapper.from_tensor`: IPC operates on the underlying storage which is always contiguous, so the check was unnecessarily restrictive for non-contiguous views

## Test plan

- [x] `import pegaflow; print(pegaflow.__version__)` succeeds
- [x] `from pegaflow.ipc_wrapper import CudaIPCWrapper` works for non-contiguous tensors (e.g. sliced/transposed views with zero storage offset)


feifei-111 (approved):
LGTM

PR #190, opened 2026-04-01
fix(connector): adapt handle_preemptions to vLLM #34805 interface change

xiaguan:
## Summary

- vLLM [#34805](https://github.com/vllm-project/vllm/pull/34805) changed `handle_preemptions` signature from `set[str]` to `KVConnectorMetadata`
- Add runtime type check in `__init__.py` to support both old and new vLLM versions
- Carry `preempted_req_ids` through `PegaConnectorMetadata` for new vLLM path

## Test plan

- [x] Deploy with new vLLM (v0.17.2rc1+) — verify preemption no longer crashes with `TypeError`
- [x] Deploy with old vLLM — verify preemption still works via `set[str]` path

### DIFF
--- a/python/pegaflow/connector/__init__.py
+++ b/python/pegaflow/connector/__init__.py
@@ -186,9 +186,15 @@ class PegaKVConnector(KVConnectorBase_V1):
         if self._worker:
             self._worker.unregister_context()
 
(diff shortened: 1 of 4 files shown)


feifei-111 (approved):
LGTM

PR #202, opened 2026-04-09
fix(connector): align hybrid save window across kv groups

xiaguan:
## Summary
- make the connector register HMA-aware metadata for KV groups
- compute hybrid save windows from the common suffix shared by all KV groups
- align worker-side save hash mapping with the scheduler window so saved blocks can seal and load correctly

## Validation
- reproduced the hybrid/SWA save mismatch on 5090 with repeated-prefix requests
- verified external prefix cache hits and successful remote loads after the fix


AethoceSora:
> verified external prefix cache hits and successful remote loads after the fix

This PR does not run properly on the latest version of vLLM. Could you please let me know which version of vLLM you used for testing?

xiaguan:
I don't recall the exact version. It's likely a vllm version around the time this PR was released — you might want to check that yourself.

PR #223, opened 2026-04-17
feat(metaserver): add heartbeat liveness and graceful Bye for prompt node cleanup

### LINKED ISSUE #222: feat(metaserver): add server heartbeat liveness and graceful Bye RPC
## Problem

When a pegaflow-server goes offline, its block hash entries in metaserver linger for up to 120 minutes (TTL). During this window, `query_prefix` returns stale nodes, causing remote fetch attempts to dead servers.

## Proposal

1. **Heartbeat** (server → metaserver, every 10s): metaserver tracks per-node liveness with a state machine: `NOT_TRACKED → HEALTHY → SUSPECT (30s) → PURGED (90s)`.
2. **Bye RPC**: on SIGTERM, server sends `Bye` for instant cleanup (skips suspect → purge).
3. **Epoch**: UUID per process start — detects restarts at the same address and purges stale entries.
4. **Query filtering**: suspect nodes are hidden from `query_prefix` results but not deleted, allowing instant recovery if the node reconnects.

## Design doc

See `docs/superpowers/plans/2026-04-17-server-heartbeat.md` in the PR.

GentleCold:
## Summary

- Add **Heartbeat/Bye RPCs** to MetaServer for server liveness tracking
- **Suspect filtering**: nodes missing heartbeats for 30s are excluded from `query_prefix` results (entries preserved for instant recovery)
- **Hard-delete purge**: nodes missing heartbeats for 90s have all block entries purged via batched single-pass `map.retain()`
- **Bye RPC**: graceful SIGTERM triggers immediate epoch-matched cleanup
- **Epoch-based restart detection**: new process at same address auto-purges stale entries from old process
- Client-side: 10s heartbeat loop with shutdown-aware `tokio::select!`, `bye()` called during server shutdown
- Liveness metrics: suspect/purged counters, active_nodes gauge

## Motivation

Previously, when a pegaflow-server went offline, its block hash entries in MetaServer lingered until the 120-minute TTL expired. During that window, `query_prefix` could route requests to dead nodes, causing transfer failures.

## Test plan

- [x] 20+ unit tests in `pegaflow-metaserver/src/store.rs` covering all liveness state transitions, sweep, purge, epoch handling
- [x] 3 integration tests in `pegaflow-metaserver/src/service.rs` (heartbeat RPC, bye purge, suspect filtering)
- [ ] `cargo test -p pegaflow-metaserver` passes
- [ ] `cargo check -p pegaflow-server` passes (full build requires RDMA libs)

Closes #222

### DIFF
--- a/pegaflow-core/src/internode/metaserver_client.rs
+++ b/pegaflow-core/src/internode/metaserver_client.rs
@@ -1,12 +1,16 @@
 use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
 
 use log::{debug, error, info, warn};
 use pegaflow_proto::proto::engine::meta_server_client::MetaServerClient as MetaServerGrpcClient;
 use pegaflow_proto::proto::engine::{
-    InsertBlockHashesRequest, NodePrefixResult, QueryPrefixBlocksRequest, RemoveBlockHashesRequest,
+    ByeRequest, HeartbeatRequest, InsertBlockHashesRequest, NodePrefixResult,
+    QueryPrefixBlocksRequest, RemoveBlockHashesRequest,
 };
-use tokio::sync::mpsc;
+use tokio::sync::{Notify, mpsc};
 use tonic::transport::{Channel, Endpoint};
+use uuid::Uuid;
 
 use crate::metrics::core_metrics;
 
@@ -28,6 +32,7 @@ impl std::fmt::Display for ClientError {
 
 const INITIAL_BACKOFF_MS: u64 = 100;
 const MAX_BACKOFF_MS: u64 = 30_000;
+const HEARTBEAT_INTERVAL_SECS: u64 = 10;
 
 pub struct MetaServerClientConfig {
     pub metaserver_addr: String,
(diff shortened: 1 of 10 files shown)


xiaguan:
The implementation no longer maintains an explicit `suspect` state machine; it only does threshold-based checks on `last_seen`. In that case, we should remove the `suspect` concept entirely from comments, README text, and metric naming instead of continuing to describe the system as if that intermediate state still exists.

Right now the wording suggests there is a first-class persisted `suspect` state in the system, but the code does not actually model one anymore. It would be clearer to rename this in terms of `healthy` / `unhealthy`, or better yet `routable` / `non-routable`, and delete the stale `suspect` terminology altogether.

xiaguan:
`Bye` currently stops the heartbeat loop, but it does not stop the registration loop, and `InsertBlockHashes` / `RemoveBlockHashes` do not carry the process-generation identifier (`epoch` in the current naming).

That means an old process can still replay delayed block registrations after `Bye`, and those entries can be written back into `map` without any corresponding liveness entry in `nodes`. Combined with `is_node_healthy()` treating untracked owners as healthy, this reintroduces a stale-owner routing path right after the graceful cleanup path that this PR is trying to add.

In other words: the lifecycle identity is only enforced for `Heartbeat` / `Bye`, but not for the registration RPCs that actually repopulate ownership state.

xiaguan:
`epoch` feels like the wrong field name here.

This code is not modeling a distributed protocol epoch / term / fencing concept. What it actually carries is a unique identifier for *this server startup*, used to distinguish different lifetimes of the same advertised node address.

A name like `startup_id` would be much easier to read, and it avoids implying stronger protocol semantics than the implementation actually provides.

xiaguan:
The purge-related metrics look semantically inconsistent right now.

`service::bye()` records `record_node_purge(1)` unconditionally, even when `store.bye()` returns `0` because the epoch does not match or the node is already absent. On top of that, `record_purge_sweep()` is also called from the `Bye` path even though both the name and description say `sweep`.

So timeout-driven purge, graceful unregister, and no-op `Bye` are currently being mixed into the same counters. That makes the metrics much harder to trust during debugging.

xiaguan:
`Bye` also feels too informal for what this RPC actually does.

This is not just a polite shutdown signal; it unregisters the current server lifetime and purges ownership state. A control-plane RPC with that kind of side effect would be much clearer with a name like `UnregisterNode` (or something equivalently explicit), so the destructive behavior is visible from the API surface instead of being hidden behind a conversational name.

xiaguan:
There is still an in-flight RPC race between `bye()` and `heartbeat()` here.

`notify_waiters()` prevents future heartbeat ticks, but it does not cancel a heartbeat RPC that has already been sent and is currently awaiting a response. So it is still possible for MetaServer to process `Bye` first and then receive a slightly earlier heartbeat for the same process generation, refreshing `last_seen` again.

The current code reduces the chance of post-shutdown heartbeats, but it does not establish the stronger guarantee that no heartbeat can take effect after `Bye` has been issued.

PR #263, opened 2026-05-12
refactor: introduce load lease reservation

xiaguan:
## Summary
- replace hash-based query/load/unpin flow with UUID load leases via ReserveLoad and ReleaseLoadLease
- remove block_hashes from LoadRequest and pass grouped lease/block-id pairs to workers
- rename core storage reservation internals away from query_prefetch/pin semantics
- update Python connector metadata and tests for lease-based load consumption
- fix metric_hll_windows CLI parsing panic seen during e2e server startup

## Breaking changes
- removes QueryPrefetch and Unpin RPCs
- LoadRequest now uses repeated LoadLeaseBlocks instead of block_ids + block_hashes
- Python EngineRpcClient uses reserve_load and release_load_lease

## Verification
- cargo check -q
- cargo test -q --no-run
- cargo test -q -p pegaflow-core --tests
- cargo test -q -p pegaflow-server --tests
- uv run pytest python/tests/test_combine_hashes.py python/tests/test_connector_fault_tolerance.py -q
- maturin develop
- uv run pytest python/tests/test_vllm_e2e_correctness.py -v -s
- prek run


xiaguan:
One thing worth tightening before merge: `LoadLease` is now an opaque, one-shot reservation handle, so `load` should not consume it before all non-mutating request validation has passed.

Today `batch_load_kv_blocks_multi_layer_inner` consumes each lease via `consume_load_lease` before checking that the lease block count matches the provided `block_ids`, and before validating layer names / registrations / slot selection. If any of those validations fails, the request returns an error but the lease ref budget has already been consumed, so the caller cannot retry the same valid reservation.

The old hash/pin flow had a similar ordering, but the lease design changes the contract: Python no longer owns the block hashes and cannot reconstruct the load intent after a failed consume. The lease should behave like a reservation that is committed only after the load target has been validated.

Suggested shape: split `load` into a non-mutating validation phase and a commit phase. First validate instance/gpu/layers/slot and check that each lease exists with the expected block count; only then consume the lease refs and submit the load task.

PR #277, opened 2026-05-15
perf(pinned-mem): NUMA-aware parallel pre-touch for pinned alloc

xiaguan:
## Summary

- Replace `cudaHostAlloc` with `mmap + parallel_pre_touch + cudaHostRegister` for the Regular allocator path; add the same pre-touch step before register on the HugePages path.
- Each touch worker calls `pin_thread_to_numa_node(node)` so first-touch places every page on the target NUMA node.
- `PinnedMemoryPool::new` / `ShardedPinnedPool::new` now plumb `NumaNode` through to the allocator; `NumaAwarePinnedPools::new` passes the per-node id, `new_global` passes `NumaNode::UNKNOWN`.
- `Drop` splits: WriteCombined → `cudaFreeHost`; Regular / HugePages → `cudaHostUnregister + munmap`.

## Why

Cold-start of large pinned pools is the dominant init cost. Measurements show `cudaHostAlloc` (the old Regular path) and the serial fault-in inside `cudaHostRegister` for HUGETLB regions are both single-threaded. Parallel pre-touch parallelizes that work and pins placement.

500GB cold start on 240-core dual-socket (h20):
- non-hugepage: **397.7s → 21.3s** (18.7x)
- hugepage:      **88.5s → 14.8s** (5.98x)

20GB on 48-core single-socket: **17.4s → 3.9s** (4.4x).

NUMA-aware vs naive (200GB, 64 threads): explicit pinning gives 1000/1000 sampled pages on the target node at no timing cost; relying on scheduler luck is non-deterministic and unsafe under concurrent multi-pool init.

See top-of-file comment in `pegaflow-core/examples/pinned_alloc_parallel.rs` for the full measurement tables and methodology.

## Test plan

- [x] `cargo build --release -p pegaflow-core --examples`
- [x] `cargo clippy --release -p pegaflow-core --examples --tests -- -D warnings`
- [x] `cargo fmt -p pegaflow-core --check`
- [x] `cargo test --release -p pegaflow-core --lib pinned_mem` (4/4)
- [x] `cargo test --release -p pegaflow-core --lib pinned_pool` (2/2)
- [x] Bench example run on h20 (500GB, hugepage + non-hugepage sweeps)
- [x] NUMA placement verified via `move_pages(2)` (separate one-off bench, not committed)

### DIFF
--- a/pegaflow-core/Cargo.toml
+++ b/pegaflow-core/Cargo.toml
@@ -60,3 +60,6 @@ harness = false
 
 [[example]]
 name = "pinned_alloc"
+
+[[example]]
+name = "pinned_alloc_parallel"
--- a/pegaflow-core/examples/pinned_alloc_parallel.rs
+++ b/pegaflow-core/examples/pinned_alloc_parallel.rs
@@ -1,0 +1,468 @@
+//! Cold-start cost of pinned memory allocation under different strategies.
+//! Motivates the `mmap + parallel_pre_touch + cudaHostRegister` path now used
+//! by `PinnedMemory::allocate_regular`/`allocate_hugepages`.
+//!
+//! Non-hugepage (size = BENCH_SIZE_GB):
+//!   A. cudaHostAlloc(flags=0)                       — old Regular path
+//!   B. mmap(MAP_POPULATE) + cudaHostRegister        — single-thread populate
+//!   C. mmap + parallel_pre_touch(N) + register      — current Regular path
+//!
+//! Hugepage (size = BENCH_HUGE_MB, requires reserved hugepages):
+//!   D. mmap(MAP_HUGETLB) + cudaHostRegister         — old HugePages path
+//!   E. mmap(MAP_HUGETLB) + parallel_touch(N) + register — current HugePages path
+//!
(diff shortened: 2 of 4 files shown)


xiaguan:
KV-Review result: approve in Slock. GitHub formal approve is blocked because this token is treated as the PR author.

Two follow-up-level notes, not merge blockers:

1. `parallel_pre_touch` currently uses `available_parallelism()` independent of allocation size. For tiny allocations this can spawn more workers than pages. A later cleanup should clamp worker count by page count and possibly keep small allocations single-threaded.
2. `pin_thread_to_numa_node` failure currently logs a warning and continues. That is acceptable as a placement degradation because the outer NUMA allocation path already runs under `run_on_numa`, but the PR wording should avoid implying every worker pin is guaranteed.

Review-side checks passed:
- `cargo check -p pegaflow-core -p pegaflow-server`
- `cargo test -p pegaflow-core pinned_mem --lib -- --nocapture`
- `cargo test -p pegaflow-core --test ssd_cache -- --nocapture`

GitHub CI is green and merge state is CLEAN.

PR #290, opened 2026-05-19
perf: add cpu path benchmarks and optimize long-block saves

xiaguan:
## Summary
- add `pegaflow-core` CPU-path Criterion benchmarks for save/query/load, metaserver enqueue, 61-layer CPU overhead, and 61-layer DtoH save
- optimize long-block query/save paths by reducing prefix key cloning, ordered multi-layer save grouping, RawBlock segment allocation overhead, and insert/read-cache batching
- align mock RPC tests with current req_id validation and wait for save visibility in test helpers

## Before / After Performance
Measured locally on the same machine with `cargo bench -p pegaflow-core --bench cpu_path`; values are Criterion medians / local point estimates from the branch work.

| Path | Before | After | Notes |
| --- | ---: | ---: | --- |
| `cpu_path/query_prefetch_lease/32768` | ~12.3 ms | ~6.1 ms | Avoids cloning prefix cache keys and skips an extra remaining-prefix clone. |
| `cpu_path/save_flush_unique/8192` | ~21.3 ms | ~13.1 ms | RawBlock 1/2 segment inline storage and save/insert allocation reductions. |
| `cpu_path/save_flush_unique/32768` | ~78.4 ms | ~55 ms | Same single-layer save path, long-block case. |
| `cpu_path/save_flush_unique_metaserver/32768` | ~125 ms | ~66 ms | Original fake metaserver path; renamed to enqueue benchmark later to make scope explicit. |
| preliminary `cpu_path/save_flush_unique_multilayer_61/8192` | ~159 ms | superseded | Old 61-layer group mixed CPU overhead with 488 MiB DtoH and shared source pointers. |

## 61-Layer CPU Path Breakdown
The final 61-layer benchmark split separates CPU slot overhead from DtoH volume:

| Path | Current |
| --- | ---: |
| `cpu_path/save_flush_cpu_multilayer_61/8192` | ~130.3 ms for 499,712 slots |
| `cpu_path/save_submit_cpu_multilayer_61/8192` | ~53.7 ms |
| `cpu_path/save_insert_flush_cpu_multilayer_61/8192` | ~77.3 ms |
| `cpu_path/save_flush_dtoh_multilayer_61/8192` | ~133.5 ms with distinct per-layer GPU sources |

## Validation
- `cargo fmt --check`
- `cargo check -p pegaflow-core --benches`
- `cargo test -p pegaflow-core internode::metaserver_client`
- `cargo test -p pegaflow-core --test engine_basic --test save_validation --test prefix_semantics --test prefetch_lease`
- commit hooks passed `cargo clippy` and `cargo test --release`; after rebasing onto latest `origin/master`, conflict resolution was rechecked with the commands above

### DIFF
--- a/pegaflow-core/src/storage/write_path.rs
+++ b/pegaflow-core/src/storage/write_path.rs
@@ -266,19 +266,18 @@ fn send_backing_batches(
     if deps.ssd_store.is_none() && deps.metaserver_client.is_none() {
         return;
     }
-  if let Some(ssd) = &deps.ssd_store {
-      let weak_blocks: Vec<(BlockKey, Weak<SealedBlock>)> = blocks
-          .iter()
-          .map(|(k, b)| (k.clone(), Arc::downgrade(b)))
-          .collect();
+    if let Some(ssd) = &deps.ssd_store {
+        let weak_blocks: Vec<(BlockKey, Weak<SealedBlock>)> = blocks
+            .iter()
+            .map(|(k, b)| (k.clone(), Arc::downgrade(b)))
+            .collect();
 
-      ssd.ingest_batch(weak_blocks);
-  }
+        ssd.ingest_batch(weak_blocks);
+    }
 
-  if let Some(client) = &deps.metaserver_client {
-      register_block_hashes(client, namespace, blocks);
-  }
-
+    if let Some(client) = &deps.metaserver_client {
+        register_block_hashes(client, namespace, blocks);
+    }
 }
 
 fn register_block_hashes(
--- a/pegaflow-core/src/storage/write_path.rs
+++ b/pegaflow-core/src/storage/write_path.rs
@@ -266,16 +266,19 @@ fn send_backing_batches(
     if deps.ssd_store.is_none() && deps.metaserver_client.is_none() {
         return;
     }
-    let weak_blocks: Vec<(BlockKey, Weak<SealedBlock>)> = blocks
-        .iter()
-        .map(|(k, b)| (k.clone(), Arc::downgrade(b)))
-        .collect();
-    if let Some(ssd) = &deps.ssd_store {
-        ssd.ingest_batch(weak_blocks);
-    }
-    if let Some(client) = &deps.metaserver_client {
-        register_block_hashes(client, namespace, blocks);
-    }
+  if let Some(ssd) = &deps.ssd_store {
+      let weak_blocks: Vec<(BlockKey, Weak<SealedBlock>)> = blocks
+          .iter()
+          .map(|(k, b)| (k.clone(), Arc::downgrade(b)))
+          .collect();
+
+      ssd.ingest_batch(weak_blocks);
+  }
+
+  if let Some(client) = &deps.metaserver_client {
+      register_block_hashes(client, namespace, blocks);
+  }
+
 }
 
 fn register_block_hashes(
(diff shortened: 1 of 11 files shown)


GentleCold:
Thanks for the PR. The performance direction looks good overall: the `RawBlock` segment specialization, ordered save grouping, read-cache batching, and MetaServer registration batching are all well-scoped optimizations for long block-list saves.

Before merging, I would like to see a couple of focused regression tests around the new fast paths:

1. **Ordered multi-slot insert fast path**

   The new ordered insert path in `process_insert_batch` / `SealedBlock::from_ordered_slot_inserts` is only indirectly covered by the single-slot case. Could we add a direct multi-slot test, for example `total_slots = 3` with slots `0, 1, 2` provided in a single entry?

   That test should assert that the block seals immediately, no inflight entry remains, the block is visible in the read cache, the footprint is correct, and the slot NUMA metadata has the expected length/value. This would protect the exact path optimized for multi-layer saves.

2. **MetaServer mixed insert/remove ordering**

   The grouped MetaServer batching path is a good optimization, but the correctness-sensitive part is the fallback to last-write-wins netting when insert/remove commands are mixed in the same drain cycle. Could we add a regression test for mixed operations on the same hash?

   Useful cases:
   - `Insert(A), Remove(A)` results in a remove
   - `Remove(A), Insert(A)` results in an insert
   - mixed operations across multiple namespaces still produce the expected per-namespace RPCs

   This would make sure the new grouped fast path preserves the previous ordering semantics.

A couple of smaller cleanup suggestions:

- The comment above `try_register_namespace` is stale: it still describes accepting a flat list of `(namespace, block_hash)` pairs, but the new API accepts one namespace plus a list of hashes. Since this is the key MetaServer hot-path optimization, the comment should reflect the new shape.
- The MetaServer benchmark uses a fake/unreachable MetaServer address and mainly measures client-side enqueue/save-path overhead, not a successful MetaServer RPC round trip. A short comment near `FAKE_METASERVER_ADDR` or the benchmark group would make that scope clearer.
- `cpu_path` requires CUDA device 0 at runtime. That matches some existing benches in this crate, so I do not consider it a blocker, but documenting the runtime requirement in the file-level comment would help future users.

Assessment: ready to merge with small fixes. The implementation looks reasonable, but the ordered insert fast path and MetaServer mixed insert/remove behavior are correctness-sensitive enough that I would prefer direct tests before merge.

xiaguan:
@GentleCold done

pegaflow-core/src/storage/write_path.rs
```diff
@@ -216,16 +274,17 @@ fn send_backing_batches(deps: &InsertDeps, blocks: &[(BlockKey, Arc<SealedBlock>
         ssd.ingest_batch(weak_blocks);
     }
     if let Some(client) = &deps.metaserver_client {
-        register_block_hashes(client, blocks);
+        register_block_hashes(client, namespace, blocks);
     }
```
GentleCold: `weak_blocks` is only used by `ssd.ingest_batch`. In a MetaServer-only configuration, including the new `save_flush_unique_metaserver_enqueue` benchmark, this still allocates and clones all block keys even though SSD is disabled.

```suggestion
  if let Some(ssd) = &deps.ssd_store {
      let weak_blocks: Vec<(BlockKey, Weak<SealedBlock>)> = blocks
          .iter()
          .map(|(k, b)| (k.clone(), Arc::downgrade(b)))
          .collect();

      ssd.ingest_batch(weak_blocks);
  }

  if let Some(client) = &deps.metaserver_client {
      register_block_hashes(client, namespace, blocks);
  }

```

GentleCold:
LGTM！

PR #292, opened 2026-05-20
fix(connector): handle split physical kv blocks

GentleCold:
## Summary
- infer the physical KV cache row layout from vLLM tensors and register PegaFlow blocks at scheduler logical-block granularity
- group split physical rows into one logical block when FlashMLA uses smaller physical blocks, e.g. physical 64 rows for logical block size 128
- keep MLA blocks-first inference while preserving the original KV-first path for non-MLA tensors
- add connector layout tests for KV-first, blocks-first, and split-physical layouts

## Test Plan
- `git diff --check`
- `cargo fmt --check`
- h20: `cd /root/kexi/pegaflow/python && maturin develop --release`
- h20: `python -m pytest tests/test_worker_kv_layout.py tests/test_combine_hashes.py` (27 passed)
- h20: `cargo test -p pegaflow-core --test prefetch_lease` (2 passed)
- h20 DeepSeek-V3.2 TP8 MLA correctness with 10GB pool:
  - command: `PEGAFLOW_POOL_SIZE=10gb bash scripts/bench/pegaflow-tp8-mla-deepseek/correctness.sh --run-id flashmla-blockfix-20260520-110956 --pool-size 10gb --num-prompts 4 --max-tokens 8 --prompt-tokens 384 --vllm-max-num-seqs 4`
  - run dir: `/root/kexi/pegaflow_bench_runs/tp8-mla/flashmla-blockfix-20260520-110956`
  - exact cold/warm outputs matched baseline; all non-partial comparisons passed
  - strict script reported one `prompt_0_partial` mismatch; treated as allowed chunk-prefill partial variance for this validation
  - external hits observed: 24 hit blocks, 64 successful load RPCs, 1.92GB loaded

### DIFF
--- a/python/pegaflow/connector/worker.py
+++ b/python/pegaflow/connector/worker.py
@@ -8,7 +8,7 @@ import threading
 import time
 from collections.abc import Iterable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import torch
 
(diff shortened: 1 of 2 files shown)


GentleCold:
request review @xiaguan

GentleCold:
image

xiaguan:
This looks good overall. One small edge case worth considering: the layout inference treats any tensor with `shape[0] == 2` as KV-first, but MLA/FlashMLA blocks-first tensors could also have `shape[0] == 2` when there is only one logical block split into two physical rows. This is probably rare in real deployments, but it may be worth either documenting the assumption or using `is_mla` to prefer blocks-first inference for MLA layouts.

xiaguan (approved):

PR #300, opened 2026-05-28
feat(connector): add save-only pegaflow mode

xiaguan:
## Summary
- Add `pegaflow.mode` extra config with `read_write` default and `save_only` mode.
- Make save-only mode skip Pega query/load while still advancing save metadata from absolute computed-token watermarks.
- Add unit coverage for save-only scheduling/resume behavior and an E2E that combines full-hit DecodeBenchConnector + Pega save-only, then verifies a read-write Pega instance can load the saved KV.

## E2E evidence
- Save-only phase metrics: `save_bytes=9175040 insertions=5 hits=0 load_bytes=0`.
- Read-write phase metrics: `hits=4 load_bytes=7340032`.
- Save-only server log contains no `cache_lookup`, `query_prefetch`, or `load`; read-write phase logs `cache_lookup: hit_blocks=4`.

## Tests
- `uv run --extra dev ruff check python/pegaflow/connector/__init__.py python/pegaflow/connector/common.py python/pegaflow/connector/scheduler.py python/tests/test_combine_hashes.py python/tests/test_vllm_save_only_e2e.py python/tests/vllm_helpers.py`
- `uv run --extra test pytest tests/test_vllm_save_only_e2e.py --collect-only -q -m e2e`
- `uv run --extra test pytest tests/test_combine_hashes.py -q`
- `PYTHONPATH=/data/pegadev/pegaflow-save-only-mode/python /data/pegadev/pegaflow/.venv/bin/python -m pytest tests/test_vllm_save_only_e2e.py -m e2e -q -s --model Qwen/Qwen3-0.6B --e2e-port 18100`
- pre-commit hooks, including `cargo test --release`

PR #298, opened 2026-05-26
refactor(metrics): use build_buckets helper for histogram buckets

jimmy-evo:
## Summary

Replace hand-written bucket lists (`TRANSFER_DURATION_BUCKETS_SECONDS`, `prefetch_buckets`, `blocks_buckets`) with a `build_buckets(mantissa_lst, max_value, start_exp)` helper that generates log-spaced `[1, 2, 4, 8] × 10^n` buckets.

Centralizes the bucket scheme across:
- `pega_prefetch_duration_seconds` — up to 2.0 s
- `pega_prefetch_blocks` — up to 2000
- `pega_load_duration_seconds` / `pega_save_duration_seconds` — up to 1.0 s
- `pega_load_blocks` / `pega_save_blocks` — up to 2000

## Test Plan

- prek hooks (ruff-check, ruff-format, isort, codespell, typos) pass
- `cargo-test-release` skipped locally (missing libibverbs headers); will be exercised by CI

PR #284, opened 2026-05-18
feat(connector): replace query pinning with leases

xiaguan:
## Summary
- Replace query_prefetch/unpin and pin refcounts with lease-backed query/release semantics.
- Collapse query results to Loading/Ready only; Ready carries num_hit_blocks and an opaque lease.
- Move core lease ownership into QueryLeaseManager with typed QueryLeaseId, instance validation, Arc<SealedBlock> ownership, release_instance cleanup, and a Tokio TTL sweeper.
- Update Python/vLLM connector flow to transfer lease ownership through scheduler -> worker load, release on cleanup/failure, and use a NoopKVConnector baseline for vLLM e2e correctness.

## Validation
- cargo check -p pegaflow-core -p pegaflow-server
- maturin develop
- python -m pytest python/tests/test_engine_client.py python/tests/test_combine_hashes.py python/tests/test_connector_fault_tolerance.py
- python -m pytest -m e2e python/tests/test_vllm_e2e_correctness.py --model <local-model-path> --tensor-parallel-size 1 --pipeline-parallel-size 1 --max-model-len 4096 -v -s
- PYTHONOPTIMIZE=1 python -m pytest python/tests/test_combine_hashes.py::TestSchedulerQueryProbeReuse::test_load_block_mismatch_releases_probe_and_raises -q
- prek run --all-files

## Notes
- Replaces the previous implementation on this PR branch with a fresh worktree implementation.
- No docs repo changes are included in this code PR.

PR #283, opened 2026-05-18
fix(server): fail on invalid rdma nics

xiaguan:
## Summary
- accept comma-separated and space-separated `pegaflow-server --nics` values, including quoted comma lists with spaces
- reject empty NIC names such as `--nics mlx5_0,,mlx5_1` instead of passing malformed values into RDMA init
- propagate RDMA transport initialization failures through engine/server startup instead of silently disabling P2P
- document both accepted `--nics` forms

Fixes #276

## Tests
- `CARGO_TARGET_DIR=/data/slock_agents/KV-Review/target-pr283 cargo test -p pegaflow-server cli_nics --lib -- --nocapture`
- `CARGO_TARGET_DIR=/data/slock_agents/KV-Review/target-pr283 cargo test -p pegaflow-server --lib -- --nocapture`
- `CARGO_TARGET_DIR=/data/slock_agents/KV-Review/target-pr283 cargo test -p pegaflow-core rdma_initialization_failure_is_returned_to_caller --lib -- --nocapture`
- `CARGO_TARGET_DIR=/data/slock_agents/KV-Review/target-pr283 cargo fmt --check -p pegaflow-server`
- `CARGO_TARGET_DIR=/data/slock_agents/KV-Review/target-pr283 cargo check -p pegaflow-server -p pegaflow-core`
- `prek run --files docs/server.md pegaflow-core/src/backing/rdma.rs pegaflow-core/src/lib.rs pegaflow-core/src/storage/mod.rs pegaflow-core/src/storage/write_path.rs pegaflow-core/tests/common/helpers.rs pegaflow-server/src/lib.rs pegaflow-server/tests/p2p_rdma.rs`
- GitHub CI green

## Notes
- `cargo test -p pegaflow-core` with default parallel integration tests hit an existing-looking `engine_basic::save_query_load_roundtrip_with_numa` flake twice (`0/4` then `2/4` hits). The same test passed alone and `engine_basic` passed with `--test-threads=1`.

PR #302, opened 2026-05-29
chore(release): bump version to 0.22.4

### LINKED ISSUE #222: feat(metaserver): add server heartbeat liveness and graceful Bye RPC

### LINKED ISSUE #276: pegaflow-server: `--nics` silently misparses comma-separated values
## Problem

`pegaflow-server --nics` only accepts space-separated NIC names (`--nics mlx5_0 mlx5_1`). Passing a comma-separated list (`--nics mlx5_0,mlx5_1,...,mlx5_7`) is parsed as a single NIC literal `"mlx5_0,mlx5_1,...,mlx5_7"` with no validation error from the CLI layer.

Downstream effect: `new_rdma` fails to open the bogus NIC, `rdma_transport` becomes `None`, `rdma_fetch` becomes `None`, and the engine **silently disables P2P remote fetch** — `try_rdma_load` short-circuits and never queries MetaServer. From the operator's perspective, MetaServer registrations look healthy but `query_prefix_blocks` count stays at zero forever.

## Why this matters

- Comma separation is the convention used by sibling flags in the same CLI (e.g. `--devices=0,1,2,3`) and is the natural form in K8s pod `args`.
- The failure mode is silent at the surface (no startup error from clap), only visible deep in pegaflow-server logs as a `Failed to initialise RDMA transport` line. Operators who skim logs will conclude P2P is enabled when it isn't.
- We just hit this in production (Kimi-K2.6 prefill pods): MetaServer had 1.3M entries and 60k+ inserts, but zero remote queries.

## Requirement

`--nics` should accept both space- and comma-separated NIC names, consistently with the other multi-value flags in this CLI. Misparsed/unopenable NICs should not silently fall back to "P2P off".

### LINKED ISSUE #294: PegaKVConnector fails to initialize with cross-layer KV cache layout
## Behavior

On current master, vLLM fails during engine initialization when `PegaKVConnector` registers a cross-layer KV cache layout.

The observed KV cache allocation shape is:

```text
(26452, 93, 2, 64, 1, 128)
```

Startup fails before the vLLM server becomes ready with:

```text
ValueError: logical block size must be a multiple of physical block size (logical=64, physical=93)
```

This happens before any benchmark or request traffic runs, so jobs that depend on the connector fail at startup.

xiaguan:
## chore(release): bump version to 0.22.4

Bumps the Rust workspace, the `pegaflow-llm` Python package, the commitizen version, and the `Cargo.lock` workspace package versions from `0.22.3` to `0.22.4`.

---

## Release notes — 0.22.3 → 0.22.4

18 PRs landed on `master` since `v0.22.3` (2026-05-15). Grouped below for release notes.

### Highlights

- **Disaggregated prefill/decode over RDMA push** (#297) — a brand-new vLLM v1 KV connector (`PdConnector`) plus a v2 RDMA transfer engine (`pegaflow-transfer/src/v2`). KV is pushed prefill→decode layer-by-layer via one-sided RDMA WRITE as each attention layer completes, overlapping transfer with the forward pass instead of pulling after prefill finishes (vLLM NIXL model). On H20 / Qwen3-8B the added TTFT is **2–4× lower than NIXL** across 512–16k input lengths.
- **Query leases replace query pinning** (#284, #288) — the query/load/release control path moved from pin refcounts to lease-backed ownership. Query results collapse to `Loading`/`Ready` only; `Ready` carries `num_hit_blocks` plus an opaque lease that transfers scheduler→worker and is released on cleanup/failure, with a TTL sweeper reclaiming abandoned leases.
- **Save-only connector mode** (#300) — new `pegaflow.mode` config; `save_only` skips Pega query/load while still advancing save metadata, so an instance can populate the cache without serving reads.

### Features

- feat(pd): RDMA push connector for disaggregated prefill/decode (#297)
- feat(connector): add save-only pegaflow mode (#300)
- feat(storage): sharded SSD cache — cache spread across multiple files, uring engine dispatches across shards, prefetch ready-blocks now ordered by requested keys (#299)
- feat(rdma): per-peer N QPs with WQE-level round-robin — new `--qps-per-peer` (default 2), round-robin at WQE level so one in-flight task saturates all QPs; handshake validates both sides agree on N (#291)
- feat(metaserver): node lifecycle fencing — heartbeat-based node tracking with per-node UUID fencing, stale nodes hidden via `--node-stale-secs` (#285, closes #222)
- feat(connector): replace query pinning with leases (#284)

### Fixes

- fix(connector): preserve non-MLA KV layout registration so cross-layer layouts (e.g. GLM-4.7-FP8) register by block stride; limit logical/physical block splitting to MLA (#295, closes #294)
- fix(numa): allocate pinned pools on GPU-local NUMA nodes instead of the full CPU NUMA set, avoiding wasted capacity on CPU-only nodes (#293)
- fix(connector): handle split physical KV blocks — group split physical rows into one logical block when FlashMLA uses smaller physical blocks (#292)
- fix(connector): allow a query lease to be consumed once per registered worker, fixing multi-worker `query lease is unknown or expired` (#288)
- fix(server): fail on invalid RDMA NICs — accept comma/space-separated `--nics`, reject empty names, propagate RDMA init failures instead of silently disabling P2P (#283, fixes #276)
- fix(connector): remove the unused scheduler pending-save request limit and save-drop accounting (#282)
- fix(connector): demote `cache_lookup_reuse` log from INFO to DEBUG to stop log spam under cache pressure (#280)

### Performance

- perf: CPU-path Criterion benchmarks + long-block save optimizations — e.g. `query_prefetch_lease/32768` ~12.3 ms → ~6.1 ms, `save_flush_unique/8192` ~21.3 ms → ~13.1 ms via reduced prefix-key cloning, ordered multi-layer save grouping, and RawBlock inline-segment allocation (#290)

### Internal / refactor / tests

- refactor(metrics): centralize histogram buckets behind a `build_buckets` helper (#298)
- refactor(core): make prefetch tasks terminal (Ready results carry RAM prefix blocks); default storage admission to no TinyLFU unless explicitly enabled (#287)
- test(server): mock vLLM gRPC E2E harness covering save/query/load/release/session contracts (#289)
- chore: tune transfer duration histogram buckets toward long-tail visibility (#281)

### Notable behavior & config changes (upgrade notes)

- **Query API**: query results are now `Loading`/`Ready` only; `Ready` exposes `num_hit_blocks` + an opaque lease. Pin/unpin refcount semantics are gone (#284, #288).
- **Release RPC**: unknown/expired leases now return `FailedPrecondition` instead of being silently accepted (#289).
- **`--nics`**: now rejects empty entries (e.g. `mlx5_0,,mlx5_1`) and fails startup on RDMA init errors rather than silently falling back to no-P2P (#283).
- **New CLI flags**: `--qps-per-peer` (default 2) (#291), `--node-stale-secs` for metaserver (#285).
- **New connector config**: `pegaflow.mode` with `read_write` (default) / `save_only` (#300).
- **Storage admission**: TinyLFU is now off unless explicitly enabled (#287).

### Full PR list

```
#298 refactor(metrics): use build_buckets helper for histogram buckets
#300 feat(connector): add save-only pegaflow mode
#299 feat(storage): add sharded SSD cache support
#297 feat(pd): RDMA push connector for disaggregated prefill/decode
#290 perf: add cpu path benchmarks and optimize long-block saves
#295 fix(connector): preserve non-MLA kv layout registration
#293 fix(numa): allocate pinned pools on GPU-local NUMA nodes
#292 fix(connector): handle split physical kv blocks
#291 feat(rdma): per-peer N QPs with WQE-level round-robin
#285 feat(metaserver): add node lifecycle fencing
#287 refactor(core): make prefetch task terminal
#289 test(server): add mock vLLM RPC E2E coverage
#288 fix(connector): allow query leases across workers
#284 feat(connector): replace query pinning with leases
#283 fix(server): fail on invalid rdma nics
#282 fix(connector): remove scheduler save limit
#281 chore: tune transfer duration buckets
#280 fix(connector): demote cache_lookup_reuse log to debug
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2279,7 +2279,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pegaflow-common"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "colored",
  "libc",
@@ -2289,7 +2289,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-core"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "ahash",
  "bytesize",
@@ -2325,7 +2325,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-metaserver"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "axum",
  "clap",
@@ -2349,7 +2349,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-proto"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "prost",
  "tonic",
@@ -2359,7 +2359,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-py"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "log",
  "pegaflow-common",
@@ -2375,7 +2375,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-server"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "axum",
  "clap",
@@ -2409,7 +2409,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-transfer"
-version = "0.22.3"
+version = "0.22.4"
 dependencies = [
  "anyhow",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.22.3"
+version = "0.22.4"
 edition = "2024"
 license = "Apache-2.0"
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "pegaflow-llm"
-version = "0.22.3"
+version = "0.22.4"
 description = "High-performance key-value storage engine with Python bindings"
 readme = "README.md"
 requires-python = ">=3.10"
@@ -110,6 +110,6 @@ profile = "black"
 
 [tool.commitizen]
 name = "cz_conventional_commits"
-version = "0.22.3"
+version = "0.22.4"
 version_files = ["pyproject.toml:^version", "../Cargo.toml:^version"]
 tag_format = "v$version"


feifei-111 (approved):
LGTM

PR #304, opened 2026-05-29
fix: add aarch64 support

hxieustc:
# Motivation

pegaflow is hard coded for x86_64. If we run pegaflow on GB200 or similar architectures, we need to add support for arm64.

# Fix

Currently pegaflow does not bind to specific architectures. The only hardware platform architecture specific bindings are (1) 64-bit pointers, (2) 64-bit atomics via AtomicI64. Both of which are supported by arm64. Thus pegaflow can run on arm64 / aarch64 platforms.

The fix is very simple and straightforward:

Instead of using 
```rust
#[cfg(not(all(target_arch = "x86_64", target_os = "linux")))]
```
We can use 
```rust
+#[cfg(not(all(target_os = "linux", target_pointer_width = "64", target_has_atomic = "64")))]
```
to allow any architecture that supports linux (shared memory) and 64-bit points and 64-bit atomic. arm64/aarch64 naturally follows in this category.

### DIFF
--- a/pegaflow-core/src/sync_state.rs
+++ b/pegaflow-core/src/sync_state.rs
@@ -1,10 +1,14 @@
 //! Shared-memory synchronization state for async KV cache loading.
 //!
-//! Only the batch-level `LoadState` is supported. Other platforms are not
-//! supported; compilation will fail outside x86_64 Linux.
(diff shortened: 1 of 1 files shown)


xiaguan (approved):
LGTM！

PR #307, opened 2026-05-31
fix: support non-RDMA local builds

ClSlaid:
## Summary
- make RDMA/pegaflow-transfer optional for core/server/python crates while preserving default RDMA builds
- keep RDMA fetch wiring behind a small storage prefetch adapter for no-RDMA builds
- split Python RDMA PyO3 bindings into `pd_rdma.rs` and pass Python loader env to bundled binaries

## Tests
- `cargo fmt --check`
- `cargo check -p pegaflow-py --no-default-features --features cuda-13 --lib --bins`
- `cargo test -p pegaflow-core --no-default-features --features cuda-13 --lib`

## Not tested
- RDMA-enabled builds and RDMA runtime paths were not tested on this machine; this environment only validated the no-RDMA CUDA build path.


ClSlaid:
Updated after Copilot review:

- Fixed `PdRdmaEngine::wait_done` to check `finished_recving` with a single lock acquisition instead of remove/reinsert.
- Added backoff to RDMA polling waits so they do not pure-spin for the full timeout.
- Removed `PYTHONHOME` from the launched binary environment and flattened env path construction.
- Split RDMA/no-RDMA `PegaEngine` methods into cfg-specific implementations and added no-RDMA tests for ignored RDMA NIC config and handshake error behavior.

I intentionally left `pegaflow-transfer/build.rs` unconditional: `pegaflow-transfer` is the RDMA transfer crate, and no-RDMA users avoid it through the optional dependency gates in `pegaflow-core`/`pegaflow-py`; making the transfer crate itself build without ibverbs would imply a separate non-RDMA transfer crate mode that this PR is not trying to support.

xiaguan (approved):
LGTM

PR #311, opened 2026-06-01
feat(ssd-cache): support multiple SSD cache paths

### LINKED ISSUE #249: feat: support multiple SSD cache paths
## Motivation

Today the SSD cache accepts a single path. For our multi-disk benchmark setup, we worked around that by creating a RAID0 array across several SSDs and passing the mounted RAID0 path to PegaFlow. That gives acceptable performance and keeps PegaFlow configuration simple.

However, some deployment environments cannot create RAID0 volumes or do not allow users to repartition, format, or manage block devices directly. In those environments, PegaFlow needs a way to use multiple SSD-backed paths without requiring external RAID setup.

## Feature request

Support configuring multiple SSD cache paths, for example:

```text
--ssd-cache-path /mnt/ssd0/pegaflow-cache --ssd-cache-path /mnt/ssd1/pegaflow-cache
```

or an equivalent config representation.

Internally, PegaFlow should distribute cache data across those paths and balance IO so users can get multi-drive throughput without relying on RAID0.

## Expected behavior

- Users can provide more than one SSD cache directory.
- PegaFlow creates and manages cache files under each configured path.
- Writes are distributed across paths instead of funneled into one device.
- Reads know where each cached block lives and fetch from the correct path.
- Capacity accounting reflects the aggregate usable space across paths.
- The placement policy should avoid magic strings and expose typed state in the implementation.

## Design questions

- Should placement be simple round-robin, hash-based striping, free-space aware, or load aware?
- Should one logical KV block live entirely on one path, or can large payloads be striped across paths?
- How should partial path failure be handled: fail fast, degrade capacity, or mark affected blocks unavailable?
- Should the public option accept repeated flags, a comma-separated list, or a structured config object?

## Why this matters

RAID0 is a reasonable operator-level workaround when the environment allows it, but it is not portable. Native multi-path SSD support would make high-throughput SSD cache deployments usable in locked-down clusters, managed Kubernetes environments, and machines where disks are exposed only as mounted directories.

Dayuxiaoshui:
Support configuring multiple SSD cache directories. Shards are distributed across paths in round-robin order so I/O is balanced across independent devices.

- Replace SsdCacheConfig.cache_path with cache_paths: Vec<PathBuf>
- Update open_cache_files to assign shards round-robin across paths
- Make --ssd-cache-path repeatable in server CLI
- Add unit and integration tests for multi-path placement
- Add io_uring availability probe so SSD tests skip gracefully in container environments that restrict the io_uring_setup syscall

Closes #249


Dayuxiaoshui:
cc @xiaguan

xiaguan:
Thanks, this is a useful direction.

One concern: `paths` and `shards` look like two different knobs. `paths` are for multiple physical SSDs without RAID, while `shards` are for multiple files on the same filesystem to drive higher throughput. The current implementation collapses both into one global `shard_count`, so with multiple paths and the default `--ssd-cache-shards=1`, only the first path gets a cache file and the other paths are not used.

Can we model this as per-path files/shards, or otherwise guarantee that each configured path gets at least one cache file?

Also, please fix the `io_uring` availability probe in the tests. It currently calls `io_uring_setup(..., NULL)`, which returns `EFAULT`, so the SSD integration tests early-return while still showing as passed.

Finally, please add a small benchmark report showing the throughput improvement from multiple paths versus a single path, with the same workload/capacity/block size. I mainly want to see write throughput and prefetch/read throughput scale when multiple SSD paths are configured.

Dayuxiaoshui:
cc @xiaguan

xiaguan:
Core SSD multi-path changes and tests LGTM now.

Please remove the benchmark and benchmark report from this PR:

- `pegaflow-core/benches/ssd_multi_path.rs`
- `docs/ssd_multi_path_benchmark.md`

The current benchmark/report does not represent real multi-SSD performance: it is based on a single backing device / synthetic setup, so the numbers can be misleading for the actual use case. I would rather merge the implementation and tests without a benchmark document that readers may over-interpret.

The benchmark can come back later once it is run on real independent SSD paths with a workload that reflects the engine prefetch path.

xiaguan:
One leftover after removing the benchmark: `pegaflow-core/Cargo.toml` still has the `[[bench]] name = "ssd_multi_path"` entry, but `pegaflow-core/benches/ssd_multi_path.rs` is gone.

This breaks the crate manifest:

```text
cargo bench -p pegaflow-core --bench ssd_multi_path --no-run
error: failed to parse manifest ... cannot find `ssd_multi_path` bench at `benches/ssd_multi_path.rs`
```

Please remove that bench stanza as well. After that, the implementation/tests still look good to me.

Dayuxiaoshui:
cc @xiaguan

xiaguan (approved):

PR #315, opened 2026-06-02
fix(connector): guard against query identity drift during prefetch Loading→Ready

xiaguan:
## Problem

When `query_prefetch` returns `QueryLoading`, the server starts a background prefetch task keyed solely by `req_id`. If the client's `computed_blocks` increases while the task is running (vLLM makes local progress), the next retry sends a different set of `remaining_hashes`. When the server returns `Ready` from the old task, `hit_blocks` can exceed `len(query_hashes)`, and the scheduler reports phantom external hits to vLLM.

## Fix

Replace `_PendingQueryProbe` with `_QueryProbe` — a flat data class where the probe's identity snapshot (`computed_blocks` + `query_hashes`) is frozen at first `Loading`:

- `hit_blocks is None` → backend still loading
- `matches()` → rejects any `Ready` whose identity drifted from the snapshot
- `mark_ready()` → enforces `hit_blocks ≤ len(query_hashes)` invariant
- `_finish_cache_lookup()` → deduplicates the result-logging path

No new dicts, no state enum — the implicit state via `hit_blocks is None` is the state machine.

## Affected paths

- `get_num_new_matched_tokens`: restructured around the new probe lifecycle
- `_release_query_probe`: skips lease release when `not probe.is_ready`
- `update_state_after_alloc`: unchanged (same `.lease` / `.leased_hashes` interface)

### DIFF
--- a/python/pegaflow/connector/scheduler.py
+++ b/python/pegaflow/connector/scheduler.py
@@ -25,16 +25,57 @@ if TYPE_CHECKING:
     from vllm.v1.request import Request
 
 
-@dataclass(frozen=True)
-class _PendingQueryProbe:
+@dataclass(slots=True)
+class _QueryProbe:
+    """One remote prefix-query snapshot.
+
+    A probe is keyed by:
+
+    * ``computed_blocks`` — blocks already computed locally when the query was
+      issued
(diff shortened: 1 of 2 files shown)


feifei-111 (approved):
LGTM

PR #317, opened 2026-06-03
feat: reject vllm register version mismatch

GentleCold:
## Summary
- Add a `client_version` field to `RegisterContextRequest` and have the PyO3 client send its Cargo package version during vLLM KV cache registration.
- Reject register requests on the server when the client and server PegaFlow versions differ, returning a `FailedPrecondition` error with both versions in the message.
- Keep the vLLM worker startup path fail-fast by surfacing register failures, and add focused unit coverage for version mismatch propagation.

## Validation
- `cargo fmt --check`
- `cargo test -p pegaflow-server --no-default-features --features cuda-12 validate_register_context`
- `cargo check -p pegaflow-py --no-default-features --features cuda-12`
- `cd python && python -m pytest tests/test_connector_fault_tolerance.py`
- h20b smoke: new `0.22.5` server rejected old `0.22.3` Python extension register request with `PegaFlow version mismatch: client=<missing> server=0.22.5`.

## Notes
- Full default Rust test was not used locally because this machine's RDMA headers fail `rdma-mummy-sys`; the targeted checks use `--no-default-features --features cuda-12`.
- Full vLLM correctness E2E was not completed on h20b because `/data/models/Qwen3-4B` is not a valid model directory there and existing Kubernetes-managed vLLM pods were running on that host. The focused register smoke covers this PR's version-mismatch behavior.

### DIFF
--- a/pegaflow-proto/proto/engine.proto
+++ b/pegaflow-proto/proto/engine.proto
@@ -23,6 +23,7 @@ message RegisterContextRequest {
   repeated uint64 kv_stride_bytes = 12;
   repeated uint32 segments = 13;
   uint32 pp_rank = 14;
+  string client_version = 15;
 }
 
 message RegisterContextResponse {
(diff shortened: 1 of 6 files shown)


xiaguan (approved):
LGTM

PR #318, opened 2026-06-03
fix(connector): skip lease release for zero-hit query probes

xiaguan:
## Problem

Users on 0.22.5 saw `query lease id must be non-empty` logged continuously.

A cache **miss** makes the server return `Ready(0)` with an **empty** lease (a lease is only minted when `hit > 0`). Since #315 the scheduler stores that zero-hit `Ready` as a `_QueryProbe` and cleans it up via `_release_query_probe`, whose guard was `if not probe.is_ready`. A zero-hit probe **is** ready (`hit_blocks == 0`, not `None`), so the guard didn't short-circuit and the connector called `release(b"")` on **every miss**. The server rejects empty lease bytes with `InvalidArgument` (`query lease id must be non-empty`), which the connector caught and logged via `logger.exception`.

**Impact: cosmetic only.** The empty lease maps to no server-side state (the server validates the bytes before any lookup), the exception is caught, and the probe is already popped — so no leak, no wrong result, no failed request. Just ERROR-log spam (and one wasted release round-trip) per miss.

## Fix

Guard on `probe.lease` instead of `probe.is_ready`. A probe holds a server-side lease iff its lease bytes are non-empty — false for both still-loading and zero-hit probes.

```python
def _release_query_probe(self, req_id, probe):
    if not probe.lease:          # was: if not probe.is_ready
        return True
    ...
```

## Tests — real vLLM + real pegaflow-server e2e, no mocks

Asserts on the server's Prometheus metrics after the real execution plan runs. Expanded beyond the single bug:

- **`test_no_data_path_rpc_failures`** — zero non-ok RPCs across **all** methods (query_prefetch/load/save/release/register), guarded by `cache_block_misses > 0` so the gate is never vacuous. Generalizes the regression: also catches a failed load/save/query.
- **`test_no_kv_load_failures`** — `pegaflow_load_failures_total == 0`. The output-equality test can't catch load failures (vLLM silently recomputes failed blocks locally, so text still matches baseline) — only the counter exposes it.

Added `fetch_pegaflow_rpc_failures()` (label-aware) to `vllm_helpers.py`.

### Reproduce → fix, verified on the GPU box
- Guard reverted (bug live): `FAILED` — `{'release/Client specified an invalid argument': 4.0}` (4 empty-lease releases rejected during the cold prompts).
- Fix applied: both new gates `PASSED`; output-match + cache-metrics still pass.
- Default unit gate: 74 passed. ruff: clean.

## Also

Switch the prek `cargo clippy`/`cargo test` hooks to `--no-default-features --features cuda-13,rdma` so they build against the CUDA 13 runtime on the dev box instead of panicking on the default cuda-12 `libcudart` symbol mismatch (`cudaEventElapsedTime_v2`).

### DIFF
--- a/prek.toml
+++ b/prek.toml
@@ -26,8 +26,8 @@ hooks = [{ id = "codespell", additional_dependencies = ["tomli"], files = '^pyth
 repo = "local"
 hooks = [
   { id = "cargo-fmt", name = "cargo fmt", language = "system", entry = "cargo fmt --all -- --check", files = { glob = ["Cargo.toml", "**/Cargo.toml", "**/*.rs"] }, pass_filenames = false, require_serial = true },
-  { id = "cargo-clippy", name = "cargo clippy", language = "system", entry = "cargo clippy --workspace --all-targets -- -D warnings", files = { glob = ["Cargo.toml", "**/Cargo.toml", "**/*.rs"] }, pass_filenames = false, require_serial = true },
-  { id = "cargo-test-release", name = "cargo test --release", language = "system", entry = "cargo test --release", pass_filenames = false, always_run = true, require_serial = true, stages = ["pre-commit"] },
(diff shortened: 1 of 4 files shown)


GentleCold:
LGTM

feifei-111 (commented):
LGTM

feifei-111 (approved):

PR #320, opened 2026-06-04
fix(tests): fail fast on vLLM startup death; align e2e cargo features; cfg-gate rdma-only query path

xiaguan:
## Summary

Three fixes that fell out of an e2e-correctness run + server log review:

1. **`VLLMServer._wait_for_ready` now fails fast when the server process dies.**
   It previously polled health endpoints for up to 600s even after the vLLM
   process had exited, turning every startup crash into a 10-minute timeout.
   `PegaFlowServer._wait_for_ready` already had this liveness check; mirror it
   and point the error at the server log file. (During validation of this PR a
   transient GPU OOM crashed both vLLM phases — the new check surfaced each
   failure in seconds with the exact log path.)

2. **The e2e fixture now builds pegaflow-server with `cuda-13,rdma`,** matching
   maturin dev builds. Previously the feature sets differed (`cuda-13` only),
   so every maturin/e2e alternation invalidated the shared cargo cache and
   recompiled pegaflow-core inside the fixture's 300s readiness window.

3. **`metaserver_client.rs` cfg-gates the rdma-only query path**
   (`ClientError`/`query_client`/`query_prefix`, consumed only by the
   rdma-gated `rdma_fetch`). Non-rdma builds emitted three dead-code warnings
   on every compile; default-feature builds (CI clippy) never saw them, which
   is why they went unnoticed. The registration/heartbeat path stays
   unconditional — `write_path` uses it in all configurations.

Also ignores `python/tests/*.log` (captured server logs tripped the typos
check via hex request IDs).

## Test plan

- `cargo check -p pegaflow-core --no-default-features --features cuda-13 --all-targets` — 0 warnings (previously 3)
- `cargo check` with `cuda-13,rdma` — clean
- `cargo clippy -p pegaflow-core -p pegaflow-server --all-targets -- -D warnings` (default features) — clean
- `cd python && uv run --extra test pytest` — 77 passed
- e2e gate: `pytest -m e2e tests/test_vllm_e2e_correctness.py --model /data/models/Qwen3-4B --max-model-len 4096` — 4 passed in 183.5s, fixture server built with `cuda-13,rdma`

Pre-existing: local `ruff format --check` (0.15.13) wants to reformat
`worker.py`/`test_connector_fault_tolerance.py` on master too — out of scope.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ venv/
 .pytest_cache/
 examples/bench_results/
 examples/pd_logs/
+python/tests/*.log
 
 # Built binaries in Python package (generated during wheel build)
 python/pegaflow/pegaflow-server
--- a/pegaflow-core/src/internode/metaserver_client.rs
+++ b/pegaflow-core/src/internode/metaserver_client.rs
@@ -3,13 +3,16 @@ use std::collections::HashMap;
 use log::{debug, error, info, warn};
 use pegaflow_proto::proto::engine::meta_server_client::MetaServerClient as MetaServerGrpcClient;
 use pegaflow_proto::proto::engine::{
-    HeartbeatNodeRequest, InsertBlockHashesRequest, NodePrefixResult, QueryPrefixBlocksRequest,
-    RemoveBlockHashesRequest, UnregisterNodeRequest,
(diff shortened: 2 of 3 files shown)


GentleCold:
LGTM

GentleCold (approved):

Issue #326, opened 2026-06-05
D2H / H2D Microbenchmark Baseline


Dayuxiaoshui:
# D2H / H2D Microbenchmark Baseline

> Isolated benchmark of GPU↔CPU pinned-memory copy throughput on a single
> NVIDIA RTX 5090.  These numbers are the foundation for a Conditional P/D
> policy that decides whether local prefill is cheaper than remote prefill.

## Purpose

P/D disaggregation moves KV cache between GPU and host memory (or between
nodes via RDMA).  Before we can write a policy layer that chooses *local*
vs *remote* prefill per request, we need to know the **local bandwidth
celling**.

This benchmark measures the raw `cudaMemcpy` throughput for:

* **D2H** – GPU → CPU pinned memory (save / offload)
* **H2D** – CPU pinned memory → GPU (load / resume decode)

It is intentionally independent of vLLM or the PegaFlow engine so that it
can be reproduced on any CUDA workstation without building the full project.

## Environment

| Component | Value |
|-----------|-------|
| GPU | NVIDIA GeForce RTX 5090 |
| GPU Memory | 32 GiB |
| CUDA | 12.8 |
| Driver | 570.x |
| OS | Linux x86_64 |
| PCIe | Gen4 x16 |

## Method

The benchmark script is `examples/bench_d2h_h2d.py`.  It constructs a tensor
that matches the KV-cache footprint of a real model, then times `copy_`
between a GPU buffer and a CPU buffer using `torch.cuda.Event`.

The default workload simulates **Qwen3-4B**:

| Parameter | Value |
|-----------|-------|
| Layers | 36 |
| Blocks / layer | 128 |
| Block size (tokens) | 16 |
| KV heads | 4 |
| Head size | 128 |
| Dtype | `float16` |
| **Total per transfer** | **144 MiB** |

Each run performs a warmup of 5 iterations followed by 50 timed iterations.
Latency is reported as p50 / p99; throughput is derived from
`bytes / elapsed_time`.

### 1. Baseline – Qwen3-4B footprint (144 MiB)

| Direction | p50 Latency | p99 Latency | Avg Throughput | Max Throughput |
|-----------|------------|-------------|----------------|----------------|
| D2H (save) | 2.65 ms | 2.68 ms | **56.9 GB/s** | 57.0 GB/s |
| H2D (load) | 2.69 ms | 2.71 ms | **56.1 GB/s** | 56.2 GB/s |

[...]

PR #327, opened 2026-06-05
feat(bench): measure native D2H/H2D transfer paths

Dayuxiaoshui:
## Summary

Rework the D2H/H2D benchmark from a standalone Torch memcpy script into the existing Rust benchmark surface under `pegaflow-core/benches`.

This removes `examples/bench_d2h_h2d.py` and updates the native `transfer_h2d` benchmark to measure PegaFlow transfer batches through `CopyDesc` and `TransferBackend`, covering:

- D2H and H2D directions
- direct vs kernel transfer backends
- contiguous, strided-device, and shuffled descriptor layouts
- multiple block-list sizes

The `cpu_path` benchmark is also extended with a load-side transfer fragmentation group, so the existing PegaEngine save/D2H path now has a matching load/H2D benchmark shape.

## Scope

This benchmark is scoped to local PegaFlow GPU<->CPU transfer/storage behavior. It does not claim to model Conditional P/D local-vs-remote policy decisions or the request-level RDMA P/D path, which also includes scheduling, request handshakes, per-layer readiness, RDMA push/wait, and decode-side completion behavior.
 https://github.com/novitalabs/pegaflow/issues/326


Dayuxiaoshui:
cc @xiaguan

xiaguan (changes requested):
Thanks for contributing this benchmark. The script itself is clean and the RTX 5090 numbers are useful as a raw CUDA/Torch memcpy reference, but I don't think this should be merged in its current form.

The main issue is that this benchmark does not measure the path it claims to inform. The PR frames the result as a foundation for Conditional P/D local-vs-remote decisions, but the current P/D path is not a local GPU<->CPU pinned-memory copy path. P-side serving pushes layer KV through the PD connector/native RDMA path (`python/pegaflow/pd_connector/prefill_worker.py` -> `RealRdmaPort.push_layer`), and D waits on RDMA completion. The important cost model there includes scheduling, handshake/open_request, per-layer readiness, RDMA push/wait, and D-side completion behavior. A standalone Torch `copy_` ceiling does not describe that decision boundary.

There is also already a closer Rust benchmark surface for the local save/load question. `pegaflow-core` has registered benches such as `pinned_copy`, `transfer_h2d`, and `cpu_path`, and the actual engine path builds `CopyDesc` batches through `PegaEngine`, pinned pools, and `TransferBackend` (`direct`/`kernel`). Those details matter: real transfers can be per-layer/per-block/per-segment, contiguous or strided, and affected by Pegaflow's allocation and worker synchronization. This script instead times one large contiguous Torch tensor copy, so it will report a hardware ceiling that is easy to reproduce but hard to apply to Pegaflow regressions or policy decisions.

Could you rework this as a Rust-side benchmark under `pegaflow-core/benches` instead of adding a standalone Python example? A useful version would extend the existing benches and measure the actual Pegaflow transfer/storage shape: direct vs kernel backend, contiguous vs strided layout, block-list size, D2H and H2D, and ideally the `PegaEngine` save/query/load path when the goal is policy evidence. For Conditional P/D specifically, I would expect this to be paired with request-level P/D trace inputs rather than using raw local PCIe bandwidth alone.

Also, please don't close the broad roadmap issue with this PR. The benchmark may address one exploratory measurement, but it does not close the PegaFlow roadmap scope.

Dayuxiaoshui:
cc @xiaguan

xiaguan:
thanks, will review it tomorrow

Dayuxiaoshui:
cc @xiaguan

xiaguan (approved):
LGTM. The rewrite now lives in pegaflow-core benches, scopes itself to local transfer/storage behavior, and the local CUDA 13 bench runs passed.

PR #332, opened 2026-06-08
fix(connector): register explicit layer ids

xiaguan:
## Summary

- move layer identity ownership to the vLLM connector: `RegisterContextRequest` now carries explicit `layer_ids`, and Python/PyO3 pass canonical IDs alongside layer names
- make the engine store by connector-declared IDs, reject out-of-range/conflicting IDs, and keep `num_layers` fixed instead of inferring/growing layer tables from registration order
- make GPU/context registration transactional across layer metadata and CUDA tensor registry rollback paths
- add a single-file pp4 P2P E2E that reproduces the old output/logprob divergence and verifies fixed RDMA load output equality

Phase 2 items intentionally stay out of this PR: cross-role namespace identity and identity-on-wire for remote slots.

## Repro / Fix Evidence

Old code (`311e0f0`, parent of this fix) on self hosted failed the local-load-vs-RDMA-load E2E after successful RDMA fetch/load:

- artifacts: `/tmp/pytest-of-root/pytest-10/test_pp4_p2p_matches_local_cac0/p2p_pp4_logs`
- local-load choice 0 text: `"的的\n\n\n\n\n\n\n"`
- RDMA-load choice 0 text: `" 0 or for identity for 0"`
- local-load choice 1 text: `" ，      "`
- RDMA-load choice 1 text: `"00000010"`
- remote RDMA evidence: `pega-remote.log:203`, `:210`, `:212`, `:220-231`

Fixed branch  passed the same pp4 P2P E2E after the final reviewer fixes:

- command: `.venv/bin/python -m pytest -m e2e python/tests/test_vllm_p2p_pp4_e2e.py --model /data/models/Qwen3-8B --pegaflow-pool-size 60gb -s -v`
- result: `1 passed in 129.65s`
- artifacts: `/tmp/pytest-of-root/pytest-15/test_pp4_p2p_matches_local_cac0/p2p_pp4_logs`
- remote RDMA evidence: `pega-remote.log:199`, `:202`, `:204`, `:214-223`
- registration order still differed across producer/local/remote, but each request carried canonical `layer_ids`

## Validation

- `./scripts/check.sh`
- `cargo test -p pegaflow-core instance::tests -- --nocapture`
- `cargo test -p pegaflow-server registry::tests -- --nocapture`
- `uv run --with pytest pytest python/tests/test_connector_fault_tolerance.py python/tests/test_worker_kv_layout.py -q`
- `uv run --with pytest --with requests pytest --collect-only -m e2e python/tests/test_vllm_p2p_pp4_e2e.py -q`
- jiuzhang 39 remote build:
  - `python -m maturin develop --release --uv --manifest-path python/Cargo.toml`
  - `cargo build --release --bin pegaflow-server-py --bin pegaflow-metaserver-py`
- jiuzhang 39 pp4 P2P E2E: `1 passed in 129.65s`
- toxic reviewer sub-agent: ready for review

### DIFF
--- a/pegaflow-core/benches/cpu_path.rs
+++ b/pegaflow-core/benches/cpu_path.rs
@@ -137,6 +137,7 @@ impl BenchFixture {
                 1,
                 1,
                 &[LAYER_NAME.to_string()],
+                &[0],
                 &[gpu.as_u64()],
                 &[total_bytes],
                 &[registered_blocks],
@@ -303,6 +304,7 @@ impl MultiLayerBenchFixture {
         let layer_names: Vec<String> = (0..MULTI_LAYER_COUNT)
             .map(|idx| format!("layer_{idx}"))
             .collect();
+        let layer_ids: Vec<usize> = (0..MULTI_LAYER_COUNT).collect();
         let layer_ptrs: Vec<u64> = match layout {
             MultiLayerGpuLayout::SharedSource => vec![gpu.as_u64(); MULTI_LAYER_COUNT],
             MultiLayerGpuLayout::DistinctLayerSources => (0..MULTI_LAYER_COUNT)
@@ -324,8 +326,9 @@ impl MultiLayerBenchFixture {
                 0,
                 1,
                 1,
-                1,
+                MULTI_LAYER_COUNT,
                 &layer_names,
+                &layer_ids,
                 &layer_ptrs,
                 &layer_sizes,
                 &layer_blocks,
--- a/pegaflow-core/src/instance.rs
+++ b/pegaflow-core/src/instance.rs
@@ -4,13 +4,7 @@
 //! multi-tenant inference instances and their associated GPU resources.
 
 use parking_lot::Mutex;
-use std::{
-    collections::HashMap,
-    sync::{
-        Arc,
-        atomic::{AtomicUsize, Ordering},
-    },
-};
+use std::{collections::HashMap, sync::Arc};
 
 use cudarc::driver::CudaContext;
 use log::info;
@@ -27,7 +21,7 @@ struct LayerMetadata {
     name_to_id: HashMap<String, usize>,
 
     /// Inverse mapping from IDs to layer names.
-    names: Vec<String>,
+    names: Vec<Option<String>>,
 
     /// GPU contexts indexed by CUDA device ID.
     gpu_contexts: HashMap<i32, Arc<GpuContext>>,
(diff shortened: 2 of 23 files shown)


xiaguan:
## Real-hardware validation

Validated both registration paths this PR touches on an 8×H200 RDMA node.

### 1. Per-layer id space — DeepSeek-V3.2 (DSA)

DeepSeek-V3.2 registers two caches per transformer layer (main MLA attention + sparse indexer, `...self_attn.attn` and `...self_attn.indexer.k_cache`) that share an in-model index. Because `PegaKVConnector` is not `SupportsHMA`, vLLM disables the hybrid KV-cache manager and **coalesces all 122 layers into a single kv-cache group** — so a group ordinal cannot disambiguate the two caches, and a group-ordinal scheme collapses `num_layers` to 61.

`fb46a56` anchors the id to the in-model index plus a name-derived intra-layer ordinal (`index * caches_per_layer + ordinal`), a dense `[0, num_hidden_layers * caches_per_layer)` space that:
- is identical on every role and PP rank — a PP rank only sees its own slice of `kv_cache_groups`, so any id based on local position would collide across ranks;
- reduces to the in-model index for conventional one-cache-per-layer models, keeping previously-stored caches addressable.

Validated on DeepSeek-V3.2 (TP=8, `--load-format dummy`): all 8 GPUs register `num_layers=122` (61 layers × 2 caches), 976 CUDA IPC tensors (122×8), completion returns, clean teardown.

### 2. PP4 + P2P RDMA — `test_vllm_p2p_pp4_e2e.py`

Ran the new PP4 P2P test (Qwen3-8B, TP1×PP4) 11 times across both registration modes:

| Path | Runs | Outputs match (RDMA load == local load) |
|---|---|---|
| per-layer (`PEGAFLOW_CROSS_LAYER_BLOCKS=0`, this PR's `_build_layer_id_space`) | 6 | 6/6 |
| default cross-layer (`layer_id = pp_rank`) | 5 | 5/5 |

Per-rank registration confirmed **disjoint global ids** under PP4 (rank0 → ids 0–8, rank1 → 9–17, rank2 → 18–26, rank3 → 27–35; `num_layers=36`). RDMA-fetched KV matched the same-server local-load output token-for-token in all 11 runs — with the old local-position scheme every rank would have claimed ids 0–8 and collided.

### Follow-up: flaky source-server assertion (not a correctness issue)

2/11 runs failed on `assert not source_failures` → `{'save/Internal error': 1.0}`. This is the consumer's best-effort tail-save (no read-only connector mode) erroring during teardown; outputs still matched every time, so no KV corruption. The test already ignores `save` failures on the **remote** consumer for exactly this reason (`ignored_methods=frozenset({"save"})`) but not on the **source** server, which also hosts a consumer. Making the source assertion symmetric would de-flake the gate (~18%, identical in both modes). Orthogonal to the layer-id change.

feifei-111 (approved):

PR #333, opened 2026-06-09
feat(core): add in-process load API returning a oneshot completion

xiaguan:
## What

Adds an in-process variant of the multi-layer KV load API that returns a
`oneshot::Receiver<Result<(), EngineError>>` instead of signalling a
caller-managed shared-memory `LoadState`.

The completion hand-off is abstracted into a `LoadCompletion` enum:

- `Shm(String)` — the existing cross-process path (the vLLM Python worker
  polls a shared-memory `LoadState`); behaviour unchanged.
- `Channel(oneshot::Sender)` — in-process Rust callers await/poll a oneshot,
  mirroring how `SaveTask` already replies.

`process_load_task` now takes `&[LayerLoadData]` (the completion is moved out
of the task before processing), and `batch_load_kv_blocks_multi_layer_inner`
takes a `LoadCompletion`. Empty loads complete immediately via `signal(Ok(()))`.

## Why

An in-process Rust embedder that registers raw device pointers and has no
second process has no `LoadState` to coordinate. The oneshot lets the embedder
poll for completion (keeping scheduler admission non-blocking) or await it.

## Validation

Exercised end-to-end against a real GPU by a downstream embedder: a
GPU→CPU→GPU byte-exact round-trip and a Qwen3-4B CPU-tier prefix restore both
drive `batch_load_kv_blocks_multi_layer_inproc`. The cross-process `Shm` path
is untouched.

---

**Stacked on #331** (`feat/page-first-block-stride`) — this branch builds on the
`block_stride_bytes` commit, so the PR is opened against that branch. Review/merge
#331 first; rebase onto `master` once it lands.

### DIFF
--- a/pegaflow-core/src/gpu_worker.rs
+++ b/pegaflow-core/src/gpu_worker.rs
@@ -18,7 +18,39 @@ use pegaflow_common::{NumaNode, pin_thread_to_numa_node};
 /// A task to load KV blocks from CPU to GPU for multiple layers
 pub(crate) struct LoadTask {
     pub layers: Vec<LayerLoadData>,
-    pub load_state_shm: String,
+    pub completion: LoadCompletion,
+}
+
+/// How a finished [`LoadTask`] hands its result back to the submitter.
+///
+/// Loads run on the GPU worker thread, so the outcome has to cross back to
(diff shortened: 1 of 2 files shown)


feifei-111 (approved):
LGTM

Issue #339, opened 2026-06-10
http_cleanup_hang_repro aborts with 'Fatal Python error: _PyThreadState_Attach: non-NULL old thread state' on master


xiaguan:
## Symptom

`pegaflow-server`'s `http_cleanup_hang_repro` integration test SIGABRTs — on a clean `master` checkout (cce4946, v0.22.8), so it currently blocks the `cargo test --release` pre-commit hook on this dev box regardless of branch.

```
$ cargo test --release --no-default-features --features cuda-13,rdma \
    -p pegaflow-server --test http_cleanup_hang_repro

running 1 test
Fatal Python error: _PyThreadState_Attach: non-NULL old thread state
Python runtime state: initialized

Current thread 0x00007bb9f1fff6c0 (most recent call first):
  File ".../site-packages/torch/cuda/__init__.py", line 638 in set_device

Thread 0x00007bb9c21ff6c0 (most recent call first):
  <no Python frame>

Extension modules: numpy..., torch._C, ... (total: 13)
error: test failed
Caused by:
  process didn't exit successfully: ... (signal: 6, SIGABRT)
```

## Environment

- master @ cce4946 (chore: bump version to 0.22.8), reproduced in a pristine `git worktree`
- rustc 1.95.0, pyo3 0.28
- torch 2.11.0+cu130 (root repo `.venv`, Python 3.13.5 from uv)
- Run env: `LD_LIBRARY_PATH=/usr/local/cuda-12.8/lib64:<uv cpython-3.13.5>/lib` (libpython needed by the PyO3-embedding test binaries; cuda-12.8 lib64 is the profile default matching the driver)
- Single 16 GB GPU box

## Analysis

`_PyThreadState_Attach: non-NULL old thread state` is CPython aborting because a thread attached a thread state while it already had one — i.e. two code paths (PyO3's GIL management in the embedded interpreter vs `torch.cuda.set_device` called during engine GPU registration) disagree about thread-state ownership. The second thread with `<no Python frame>` suggests the cleanup/registration race the test was written to reproduce now trips a CPython invariant before reaching the original hang scenario.

Likely interactions to investigate:

- pyo3 0.28 attach/detach semantics vs Python 3.13 (the `_PyThreadState_Attach` hard-abort is new-ish CPython behavior; 3.12 tolerated more).
[...]

Issue #338, opened 2026-06-10
E2E harness: fail fast on client/server version mismatch instead of a buried log 3 minutes in


xiaguan:
## Problem

The vLLM E2E gate depends on two artifacts that must come from the same source tree, but nothing checks this before spending minutes on startup:

1. `target/release/pegaflow-server` — picked up by `python/tests/conftest.py::find_server_binary` (the installed-package lookup typically fails with `ImportError` in dev venvs, so the cargo target binary wins).
2. The `pegaflow` extension imported by the spawned `vllm serve` process — whatever is installed in the active venv, which may be a stale wheel.

When they diverge, the run fails ~3 minutes in, at vLLM server startup, and the actual cause is buried in a log file:

```
RuntimeError: vLLM server exited with code 1, see /tmp/pytest-of-.../e2e_logs0/pegaflow.log
```

with the real error only inside that log:

```
pegaflow.PegaFlowError: register_context_batch RPC failed: code: 'The system is not in
a state required for the operation's execution', message: "PegaFlow version mismatch:
client=0.22.5 server=0.22.8"
```

Hit on 2026-06-10: rebuilt the server binary from a feature branch (0.22.8) while the dev venv still had a 0.22.5 wheel installed. All 4 E2E tests errored at setup; recovery required knowing to run `maturin develop` into the right venv (the *root* repo venv that `vllm serve` uses, not the `python/.venv` that pytest itself runs in — an extra trap).

## Proposed fix

Preflight check in the E2E fixtures, before spawning anything:

1. Resolve the server binary (existing `find_server_binary()` logic) and get its version, e.g. via `pegaflow-server --version`.
2. Compare against the client extension version (`pegaflow.__version__` or equivalent) **as resolved by the interpreter that will run `vllm serve`**, not the pytest interpreter.
3. On mismatch, fail immediately with an actionable message, e.g.:

```
PegaFlow version mismatch before E2E startup:
  server binary : 0.22.8  (target/release/pegaflow-server)
  client wheel  : 0.22.5  (/data/.../.venv/lib/python3.13/site-packages/pegaflow)
Rebuild the client into the venv vLLM uses:
  cd python && maturin develop --uv --no-default-features --features cuda-13,rdma
```

Optional hardening, independent of the preflight:

- `VLLMServer._wait_for_ready` already knows the log path on failure; tail the last N lines (or grep for `PegaFlowError|panicked`) into the raised `RuntimeError` so setup failures are self-explanatory without opening the log.
[...]

PR #340, opened 2026-06-10
refactor(core): replace KVCacheRegistration with validated KVCacheLayout

xiaguan:
## Why

`KVCacheRegistration` was an 11-public-field bag whose layout semantics were re-derived at every consumer:

- The `segments == 2 && kv_stride > bytes_per_block` is_split predicate was hand-copied in **three places** (load path, save path, pinned allocator).
- Host-side block pointers were computed **twice** with two hand-written copies of the same offset formula: once for GPU copy destinations (Phase 2), once for deferred `RawBlock` construction in the insert worker (Phase 4) — these must agree for correctness, and nothing enforced that.
- Validation was spread across runtime checked-math on the hot path instead of happening once at registration.

## What

**New `layout` module** — `KVCacheLayout` with private fields:
- All validation at construction; split-vs-contiguous is normalized into a `SegmentLayout` enum (invariant in the type, not in runtime re-checks).
- `block_copies(block_idx)` is the single GPU-addressing question the copy paths may ask. `transfer::segment_offset` and `gpu_worker::device_addr` are absorbed; the hot path is plain arithmetic, no checked ops.
- Constructor takes `segment_bytes` (not the misleading `bytes_per_block`) — for K/V split layouts the RPC field `bytes_per_block` is actually one segment, and passing a whole block would mostly pass validation with every address wrong.
- Previously-silent broken configs are now rejected at registration: overlapping `kv_stride < segment_bytes`, split with >2 segments, contiguous multi-segment with overlapping block stride.

**`RawBlock` constructed once, at allocation time** — the save path builds `Arc<RawBlock>` in Phase 2 and shares it with the GPU copy task:
- `LayerAlloc`, its `block_dst`/`make_raw_block` pair, and the blockwise-vs-batch `(alloc_idx, offset_in_alloc)` index dance are deleted — host pointers are computed exactly once.
- The insert worker is now a pure group-by-hash over prebuilt blocks.
- Load and save share one data shape (`TransferBlock` / `LayerTransferData`) and one `build_copy_descs`; `SaveBlock`'s two `unsafe impl Send` are gone (`RawBlock` is already `Send + Sync`).

**Crash early over silent degradation**:
- `service.rs` rejects negative `block_ids` at the RPC boundary; the engine rejects out-of-range block ids instead of silently skipping them; the engine-level length-mismatch check stays (in-process callers bypass RPC validation).
- Redundant downstream invariant re-checks removed (slot range checks in `ensure_all_slots_registered` and the save path, `unreachable!()` re-matches).

Net: **-200 lines**, `instance.rs` down to topology-only concerns, one source of truth each for GPU-side and host-side addressing.

## Verification

- `cargo test --release --no-default-features --features cuda-13,rdma` — full workspace green except `http_cleanup_hang_repro`, which aborts identically on master (pre-existing, tracked in #339).
- `cargo clippy --workspace --all-targets` with `-D warnings` — clean.
- vLLM E2E correctness gate: `pytest -m e2e tests/test_vllm_e2e_correctness.py --model Qwen3-4B --max-model-len 4096` — **4/4 passed** (baseline match, cache metrics, no data-path RPC failures, no KV load failures).

Reviewer note per repo convention: please rerun the E2E gate on the GPU machine. Related tooling friction filed as #338 (E2E version-mismatch preflight).

### DIFF
--- a/pegaflow-core/src/gpu_worker.rs
+++ b/pegaflow-core/src/gpu_worker.rs
@@ -5,19 +5,17 @@ use log::{debug, error, info, warn};
 use logforth::diagnostic::ThreadLocalDiagnostic;
 use tokio::sync::{mpsc, oneshot};
 
+use crate::EngineError;
 use crate::block::RawBlock;
+use crate::layout::{BlockCopies, KVCacheLayout};
 use crate::metrics::core_metrics;
-use crate::pinned_pool::MappedPinnedPtr;
 use crate::sync_state::LoadState;
-use crate::transfer::{
-    self, CopyDesc, KernelBackend, MemcpyBackend, TransferBackend, TransferMode,
-};
-use crate::{EngineError, KVCacheRegistration};
+use crate::transfer::{CopyDesc, KernelBackend, MemcpyBackend, TransferBackend, TransferMode};
 use pegaflow_common::{NumaNode, pin_thread_to_numa_node};
 
 /// A task to load KV blocks from CPU to GPU for multiple layers
 pub(crate) struct LoadTask {
-    pub layers: Vec<LayerLoadData>,
+    pub layers: Vec<LayerTransferData>,
     pub completion: LoadCompletion,
 }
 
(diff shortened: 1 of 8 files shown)


GentleCold:
I think there is still a correctness hole in the new `KVCacheLayout` validation.

`KVCacheLayout::new()` defaults `block_stride_bytes` to `segment_bytes`, and `check_bounds()` only verifies that the last addressed byte fits inside `size_bytes`. It does not verify that the address ranges for different blocks/segments are disjoint.

Two concrete cases still pass construction but produce overlapping GPU copy ranges:

```rust
// Split layout: block 0 V overlaps block 2 K.
KVCacheLayout::new(0x1000, 200 * 1024, 100, 1024, 2048, 2).unwrap();

// Adjacent K/V is normalized to Contiguous, but the default block stride
// remains one segment, so block 0's 2-segment copy overlaps block 1.
KVCacheLayout::new(0x1000, 200 * 1024, 100, 1024, 1024, 2).unwrap();
```

After this refactor, `block_copies()` trusts construction-time validation and uses plain arithmetic, so accepting either layout can make save/load copy the wrong KV bytes. This also contradicts the PR description that overlapping `kv_stride` configs are rejected at registration.

I think the constructor/bounds check needs a stronger range-disjoint invariant. For split layouts, at minimum the K region and V region should not overlap, e.g. `kv_stride_bytes >= (num_blocks - 1) * block_stride_bytes + segment_bytes` for the current affine layout. For adjacent multi-segment contiguous layouts, either default `block_stride_bytes` should be `block_bytes()` or registration should reject the layout unless an explicit stride large enough for the full contiguous block is provided.

xiaguan:
Fixed the KV cache layout overlap validation; please review.

GentleCold (approved):
LGTM

PR #342, opened 2026-06-10
refactor(transfer)!: replace v1 engine with v2 push-based transfer

xiaguan:
## What

Removes the v1 transfer engine entirely and promotes v2 to the crate root of `pegaflow-transfer`. Cross-node KV fetch is inverted from requester-driven RDMA READ to a push-based `PushBlocks` protocol: the requester allocates per-NUMA pinned slabs and sends its MR table + per-slot destination addresses; the holder scatter-WRITEs directly into requester memory and the RPC response doubles as the completion signal (RC WRITE completion implies remote placement).

Breaking: no compat shims or aliases, no v1/v2 rolling-upgrade wire compatibility.

## Review doc

Full change record (deleted features, behavior changes, both toxic-review rounds and fixes, follow-ups) lives in the docs repo: `docs/projects/transfer-engine-v2-migration.md`.

## Highlights

- **Engine**: WRITE-only data plane (Single/Paged/Scatter/Imm/Barrier), UD control plane with lazy per-peer RC handshake, one busy-poll worker thread per domain group (1..=8 NICs).
- **PushBlocks error contract**: `PushBlocksError::Rejected` (no WRITE submitted → `FAILED_PRECONDITION` → requester recycles slabs) vs `Failed` (WRITEs may be in flight → `INTERNAL` → requester leaks slabs, tracked by `rdma_fetch_leaked_bytes`). Leak-vs-recycle is encoded in the type.
- **Handshake reliability**: 100 ms retransmit driven from `poll_progress` (zero cost when idle), `is_reply` request/reply discipline (requests always idempotently re-answered, replies never answered), `rc_to_reset()` before QP activation so re-entry after partial failure is safe.
- **Shard math**: last byte-shard takes the remainder; lengths > `u32::MAX` rejected at submit (32-bit SGE).
- **Engine resilience**: cancelled callers no longer stop the engine; worker completions bail on stop signal instead of blocking.

## Validation

- `./scripts/check.sh` + full prek (fmt, clippy, typos, `cargo test --release`) green.
- Real RDMA on 8x H200 / 400G IB: `pegaflow-cpu-bench` ~368 Gbps per NIC, p50 13.7 ms for 150x4 MB tasks — UD handshake, retransmit, and scatter WRITE exercised on hardware.
- `p2p_rdma` ignored IT (engine A save → metaserver discovery → engine B RDMA fetch → integrity check) being run on an internal 8-GPU RDMA node; result will be posted in a comment.
- New `scripts/bench_p2p_load.py`: end-to-end TTFT comparison of cold prefill vs RDMA p2p fetch vs local warm hit (two servers + two vLLM replicas on one node).


xiaguan:
p2p_rdma ignored IT run on an internal 8-GPU RDMA node (8x H200, 400G IB, mlx5_1): **passed** in 9.12s after 5f58fc1 (the test needed the multi-thread tokio flavor because the v2 engine uses `block_in_place`). Save on engine A → metaserver discovery → engine B RDMA push fetch → GPU data integrity all verified on hardware.

xiaguan:
P2P load bench on an internal 8-GPU RDMA node (8x H200, 8x 400G IB, Qwen3-8B TP8, two pegaflow servers + two vLLM replicas sharing the GPUs; `scripts/bench_p2p_load.py`):

| prompt tokens | cold TTFT | p2p TTFT (before → after) | warm TTFT | RDMA fetched |
|---|---|---|---|---|
| 2k | 40ms | 256 → **137ms** | 44ms | 282MiB |
| 4k | 52ms | 460 → **161ms** | 47ms | 564MiB |
| 8k | 88ms | 921 → **255ms** | 66ms | 1124MiB |
| 16k | 179ms | 1792 → **447ms** | 85ms | 2242MiB |
| 31k | 372ms | failed¹ → **872ms** | 136ms | 4357MiB |

¹ at ~2000 blocks the QueryBlocksForTransfer response exceeded tonic's 4MiB client decode default and the fetch silently fell back to recompute.

What changed (commits `ac9b283`..`247c367`):
- **Segment coalescing**: TP8 slices each block into 576 ~4KiB segments; the push was WR-rate-bound (~1.3GiB/s). Holder pinned storage keeps consecutive *blocks* adjacent per layer (split K/V), so both ends now iterate slot-major — measured `segments=143424 coalesced=~350` per push (~400×).
- **mr_index instead of per-segment descriptor clones**, last-hit region cache.
- **64MiB internode gRPC message limits** (fixes the 31k failure).
- Holder-side `PushBlocks timing` breakdown log.

At 31k the RDMA stage moves 4.3GiB in ~100ms (near line rate for the NUMA-local NIC groups); the remaining p2p TTFT is a ~100ms fixed overhead (engine fixed latency ~85ms + gRPC ~30ms transport of per-slot metadata) — both are follow-up targets (RPC slimming to (src, dst, size) runs; per-peer multi-QP / fixed-latency hunt in the engine).

Note: cold beats p2p here because Qwen3-8B TP8 prefill is extremely cheap (~12ms/1k tokens); the bench's purpose is measuring the fetch path itself, and the crossover favors P2P for prefill-heavy models.

xiaguan:
Bandwidth ablation on the RDMA stage (holder `PushBlocks timing`, first-push ~90ms handshake excluded):

| push size | baseline (4 NICs, byte-shard) | +NUMA affinity (8 NICs) | +whole-target round-robin (`7b08d52`) |
|---|---|---|---|
| 0.55 GiB | 23-26 GiB/s | 26 GiB/s | **43 GiB/s** |
| 2.2 GiB | 41-42 GiB/s | **67-75 GiB/s** | 63 GiB/s |
| 4.26 GiB | 40-43 GiB/s | 67-82 GiB/s | **73-82 GiB/s** |

- The bench previously passed `--disable-numa-affinity` (copied from the e2e harness), which left the holder pool in one region on one engine — half the NICs idle. Removing it engages both NUMA engines (~1.8×).
- `AllDomainsShardPeers` keeps each ~2MB coalesced run whole on one NIC instead of splitting it N ways; clearly better for small pushes, neutral for large — kept.
- Current plateau is ~80 Gbps per NIC vs ~370 Gbps that `pegaflow-cpu-bench` reaches on a single NIC: the next lever is per-peer multi-QP / WR pipelining in the engine (single RC QP per peer per NIC today; measured data says 400G large messages need ≥2 QPs).

xiaguan:
## Worker-pinning ablation: exact-core pin vs NUMA-set affinity

Per-write-op instrumentation (73a8ccc) exposed two scheduling pathologies in the push path, both caused by exact-core pinning when two pegaflow-server processes share one machine (the loopback bench topology): each process deterministically pins its busy-poll RDMA worker to the **same** core (last CPU of the NUMA group). Two spinners time-slice one core (~12ms CFS slices each, observed at ~50% CPU), so:

- the second NUMA group's scatter op sat **15.5ms** in the submit channel before its worker woke up;
- after the last WR was posted the worker got descheduled while the NIC streamed, leaving CQEs unpolled for **~12ms**.

Fix (24c4c2a): restrict the worker to the NUMA node's CPU set instead of one core. The scheduler then spreads competing workers onto idle cores (observed: 4 workers on 4 distinct cores, 100% each). This also removes the need to know per-machine core counts.

**16k tokens, Qwen3-8B TP8, ~2.2GiB per push (8×400G IB, single-host loopback):**

| metric | exact-core pin | NUMA-set affinity |
|---|---|---|
| rdma_ms (submit → all complete) | 39.0 | **24.1** |
| inter-group submit stagger | 15.5ms | 0.3ms |
| completion-detection tail | ~12ms | ~0.1ms |
| wire time / aggregate during window | 11.7ms / 187GiB/s | 11.8ms / **187GiB/s** |
| per-NIC streaming | 23.7GiB/s | 23.5GiB/s |

Per-NIC streaming is unchanged — it is a **platform wall, not an engine limit**: `pegaflow-cpu-bench --memory host` does 43GiB/s on one NIC alone, but 4 concurrent NICs on one NUMA node drop to 24–30GiB/s each. In loopback every byte is DMA-read and DMA-written on the same sockets (~190GB/s mixed DRAM traffic/socket at the observed rate), so host memory bandwidth caps the test. On two separate hosts each socket carries only one direction; the same engine extrapolates to roughly 2× per NIC (~300GiB/s+ aggregate for 8 NICs).

With scheduling fixed, holder-side CPU now dominates rdma_ms: ~12ms coalescing loop over 575k proto segments + build_ms 24–32ms — exactly what the planned RPC slimming (slab base + implicit offsets instead of per-segment addrs) removes.

xiaguan:
## Hard evidence: the per-NIC plateau is the socket's DRAM bandwidth

Follow-up to the ablation comment — the "memory bandwidth wall" claim is now measured, not inferred. Method: read the IMC CAS counters (`perf stat --per-socket -a -e uncore_imc/event=0x5,umask=0xcf/ -e .../umask=0xf0/`, counts × 64B) while each workload runs on socket 0 of a 2× Xeon Platinum 8558 box.

| scenario | S0 DRAM read | S0 DRAM write | total | bench result |
|---|---|---|---|---|
| idle | 0.05 GB/s | 0.05 GB/s | ~0.1 GB/s | — |
| **platform peak**: 48-thread OpenMP triad, NUMA-local | 132.3 GB/s | 72.7 GB/s | **205.0 GB/s** | triad 220 GB/s app-level |
| 1-NIC host loopback (cpu-bench) | 46.2 GB/s | 46.2 GB/s | 92.5 GB/s | 43.0 GiB/s |
| **4-NIC host loopback, concurrent** | 103.0 GB/s | 102.7 GB/s | **205.7 GB/s** | 23.9 GiB/s × 4 |

The 4-NIC run drives the socket to exactly the triad-measured peak (205.7 vs 205.0 GB/s, +0.3%), and per-NIC throughput settles at peak/8 (each byte is DMA-read and DMA-written on the same socket in loopback). PCIe is ruled out by topology: each 400G NIC has its own PCIe switch and x16 uplink (NIC↔GPU is PIX, NIC↔NIC is NODE in `nvidia-smi topo -m`), and full-duplex Gen5 x16 carries 46 GB/s per direction comfortably.

So: the transfer engine drives the NICs to whatever the memory system allows; in this single-host bench topology the ceiling is DRAM, reached. A cross-host deployment splits the read side and write side onto different machines' sockets, removing the 1:1 read+write overlap that caps this test.

xiaguan:
## Slim RPC + build-path optimization results

Two commits since the pinning fix:

1. **`1484339` perf(transfer)!: slim transfer RPCs** — Query/Push no longer carry per-segment address tables (3MB + 6MB at 16k tokens). The query returns a per-slot size/NUMA template + block count; the requester bump-allocates destinations slot-major from per-NUMA slabs and sends only the slab bases (~3KB + ~300B, token-count independent). The holder replays the identical allocation order and rejects on any cursor divergence before submitting a WRITE. The p2p IT verifies byte-identical data content end-to-end.
2. **`c83b5de` perf(core): hoist slab cursor lookups and drop per-segment rebuild pass** — fixes the holder-side regression the stage logs exposed (per-segment HashMap lookups + 2 Arc clones × ~575k segments), and builds `Arc<RawBlock>` inline during bump allocation so the requester's post-push rebuild pass collapses into `SealedBlock::from_slots`.

### 16k-token p2p TTFT (Qwen3-8B TP8, 2×400G-class loopback, median of trials)

| | v1 baseline | NUMA-pin fix | slim RPC | + build-path opt |
|---|---|---|---|---|
| p2p TTFT | 454ms | ~310ms | ~256ms | **185ms** (170/200) |
| cold TTFT (reference) | ~178ms | ~178ms | ~178ms | 178ms |

p2p is now ~1.0× cold prefill at 16k tokens (warm hit: 93ms).

### Stage breakdown (holder / requester, ~2.2GiB per fetch)

| stage | slim RPC | + build-path opt |
|---|---|---|
| holder build (replay) | 42.7–46.1ms | **25.0–25.3ms** |
| holder rdma | 23.9–24.1ms | 24.0–35.5ms |
| requester query | 13.2ms | 15.2–23.9ms |
| requester build | 28.5ms | 25.2–27.2ms |
| requester push RPC | 67–91ms | **49.8–61.5ms** |
| requester rebuild | 54–82ms | **1.1ms** |

Remaining time is dominated by the RDMA transfer itself, which sits on the loopback DRAM wall documented above.

xiaguan:
## Data-plane optimization results (follow-up to the bandwidth regression report)

The reported gap (v2 data plane ~20 GB/s vs v1's ~89 GB/s) was CPU-bound per-segment processing, not RDMA: at 36 layers × TP8 × 1000 blocks the fetch handles 576k segments, and per-segment work dominated the wire time. Five commits attack that:

| Commit | Change |
|---|---|
| `e015d20` | Holder replay run-merge (576k → ~576 PushSegments) + requester run-granular slab allocation |
| `738b50d` | Zero-alloc homogeneity scan on the holder query path |
| `6edae96` | NUMA-partitioned sealed-block build (kills Arc refcount contention) + parallel scan |
| `8d174b9` | Seal fetched blocks from template metadata; fetched blocks now carry `slot_numas` and are re-servable |
| `2a8b830` | Overlap requester sealed-block build with the PushBlocks RPC |
| `ef5d249` | Parallel holder replay with per-run end checks + adaptive scan thread count |

### Benchmark

New end-to-end bench/test: `pegaflow-server/examples/p2p_bench` — real gRPC + MetaServer + CUDA IPC save path on the holder, real fetch path on the requester, byte-level verification of every block on every TP rank. Workload: 36 layers, TP8, 4 KB K/V segments, 1000 blocks = 2250 MiB per set.

**E2E fetch median (warm), 2250 MiB/set:**

| Setup | before (`c83b5de`) | after (`ef5d249`) |
|---|---|---|
| Single node loopback, 8 NICs | 88.9 ms (24.7 GiB/s) | **22.0 ms (99.7 GiB/s)** |
| Two nodes, 8×400G IB | — | **22.0 ms (99.9 GiB/s)** |

**Cross-node stage breakdown (after):** query 3.3 ms · holder replay 2.0 ms · RDMA WRITE 9.4 ms (2.2 GiB ≈ 250 GiB/s aggregate over 8 NICs) · requester sealing 15.5 ms fully overlapped with the push · rebuild 0.06 ms.

Byte verification passes on all sets in both setups, and the `p2p_rdma` integration test passes on the final commit.

Versus the original measurements (1.345 GB warm fetches): master was 151.7 ms e2e (8.9 GB/s) and the earlier v2 was 106.0 ms (12.7 GB/s); the data-plane regression is gone and e2e is now ~11× faster than master.

xiaguan:
**Follow-up:** real-workload measurement (MLA model, warm phase dominated by single-block ~6 MiB fetches) caught a regression the 1000-block benchmark above could not see: the replay/sealing thread fanout was unconditional, so a 1-block push paid ~0.5 ms of thread spawns for microseconds of work — small warm fetches roughly doubled end to end.

Fixed in `1047b7d`: thread count now scales with total work (blocks × slots), running inline below 64k visits. Validated on both shapes:

| Shape | before fix | after fix |
|---|---|---|
| 1000-block set (2250 MiB) | 22.0 ms e2e | 22.0 ms e2e (unchanged) |
| 1-block fetch push_ms (loopback) | ~1.1 ms | **0.33–0.53 ms** |

The small-fetch shape is now part of the validation matrix.

PR #345, opened 2026-06-11
feat(pd): seal handshake wire schema in pegaflow-pd-wire

xiaguan:
## Summary

Seal the P/D KV-push handshake (the D -> P payload) as an executable contract:

- New dependency-light crate `pegaflow-pd-wire` (serde only) defining the wire types — `Handshake` / `Layer` / `Region` / `MrDesc` — plus the imm encoding (`fail = imm ^ 0x8000_0000`, `abort = imm ^ 0x4000_0000`, top two bits reserved) and `validate()` covering the schema invariants (strictly increasing block ids, ordered regions, distinct imm ids, unique layer_idx, compact block-id resolution).
- `PdRdmaEngine.register_remote` now consumes the wire JSON string and parses it through the crate, replacing the hand-rolled PyDict traversal. The boundary between Python connector and native engine now speaks exactly the sealed format, so an external engine consuming this crate is guaranteed wire-compatible.
- Failure atomicity fix: all fallible parsing/validation now happens before any imm/counter registration (the old code could leave a stale `imm_to_req` entry when layer parsing failed).
- Tightened: duplicate `layer_idx` and imm ids colliding with the fail/abort flag bits are now rejected; the decode imm allocator wraps at `0x3FFF_FFFF` so allocated ids always keep the reserved bits clear.

Python-side dataclasses in `metadata.py` are intentionally untouched (phase 2: expose the wire types via PyO3 and delete the hand-written to_dict/from_dict).

## Validation

- `cargo test -p pegaflow-pd-wire`: 6/6, including a fixture that mirrors the exact JSON `handshake_to_compact_dict` + `json.dumps` emits (explicit nulls included).
- `python/tests/pd`: 117 passed.
- prek hooks (cargo fmt/clippy `-D warnings`/release test with `cuda-13,rdma`, ruff, isort, typos): all green.

## Known follow-ups (from self-review)

- `pd_imm()` request-id-hash fallback (used only when a handshake omits `imm_id`; the in-tree decoder always sends one) can produce ids with the reserved flag bits set — should be masked with `MAX_IMM_ID` or the fallback removed.
- `validate()` does not yet cover `MrDesc` contents (empty `addr_rkey_list`, zero `ptr`); that check still lives in the binding's `DomainAddress` parse.
- The imm flag constants are still duplicated in `pd_rdma.rs` and as Python literals; Rust-side dedup is trivial, Python side lands with phase 2.


GentleCold:
I think this should be fixed in this PR rather than left as a follow-up, because it leaves the sealed wire contract internally inconsistent.

The new wire schema reserves the top two bits of `imm_id` for fail/abort signaling and validates this when `imm_id` is present. However, `imm_id` is still optional in the schema. When it is omitted, `PdRdmaEngine::register_remote` falls back to `pd_imm(&remote_request_id)`, and that fallback returns an unrestricted 32-bit hash. It can therefore produce an id with `0x8000_0000` or `0x4000_0000` already set, which violates the same `MAX_IMM_ID` rule that the wire crate is trying to enforce.

That matters especially because this PR's goal is to make the D -> P payload an executable contract for external engines, not just the current in-tree Python decoder. The current decoder always sends `imm_id`, but an external producer can legally omit it according to the new schema, and then native may synthesize an invalid base imm.

Could we either make `imm_id` required in the wire contract, or mask the fallback value before deriving fail/abort immediates? For example:

```rust
let imm = handshake
    .imm_id
    .unwrap_or_else(|| pd_imm(&remote_request_id) & pd_wire::MAX_IMM_ID);
```

This keeps the base `imm_id` in the low 30-bit namespace and preserves the `fail = imm ^ 0x8000_0000` / `abort = imm ^ 0x4000_0000` encoding invariant.

xiaguan:
@GentleCold Fixed in b908ae6 — went with the stronger option: `imm_id` is now a required `u32` in the wire contract instead of masking the fallback.

- `pegaflow-pd-wire`: missing or null `imm_id` is rejected at parse time; `validate()` additionally rejects explicit `fail_imm_id`/`abort_imm_id` overrides with both flag bits clear, so signal immediates can never alias another request's completion imm.
- `register_remote`: the `pd_imm()` hash fallback is gone (it was an implicit out-of-band contract — exactly what the sealed schema should eliminate; the in-tree decoder always sends `imm_id` anyway).
- `wait_done` on an unregistered request now raises immediately instead of waiting 30s on hash-derived imms nobody sends, matching `fail_request`/`abort_request`.
- Python `handshake_from_dict` requires `imm_id` too, so an external producer omitting it fails at the parse boundary on P. Tightening `PdHandshake.imm_id` itself to `int` lands with phase 2 when the dataclasses are replaced by the PyO3-exposed wire types.

`cargo test -p pegaflow-pd-wire` (7), `tests/pd` (117), clippy/fmt/prek all green. Please review again.

GentleCold (approved):

Issue #344, opened 2026-06-11
Discussion: engine-derived topology fingerprint for namespaces


xiaguan:
## Context

Since the seal-based registration contract (fix/mtp-layer-count), the engine derives the instance layer-id space from the layers workers actually register: sorted layer names → dense ids, slot ownership validated at seal time. Blocks are stored keyed by `(namespace, hash)` with a slot array of `num_layers * tp_size` entries.

The namespace is still entirely client-chosen. Two instances whose layer sets differ — the obvious case is the same model with MTP enabled vs disabled (`num_nextn_predict_layers` present in config, drafter layers registered or not) — can legally point at the same namespace. Their blocks are mutually unusable:

- Different slot counts: load now fails loudly (`stored block has N slots but instance expects M`), which is correct but means a warm namespace keeps reporting hits that can never be served.
- Same slot count, different layer sets or geometry: undetectable today. The block decodes "successfully" into the wrong layers.

We currently operate under the assumption that a namespace is only ever used by one KV layout (deployment discipline). This issue is about whether the engine should enforce that instead.

## Proposal

At seal time the engine has everything that defines block layout compatibility:

```
fingerprint = hash(sorted (layer_name, bytes_per_block, segments))[:8]
namespace_final = f"{client_namespace}.{fingerprint}"
```

Appending the fingerprint server-side would make incompatible layouts land in disjoint namespaces by construction — MTP on/off, different quantization (bytes_per_block), different attention layouts all separate automatically, with zero client coordination.

Open questions:

1. `block_size` (tokens per block) is part of compatibility but not visible to the engine at registration — it lives in the connector's hash computation. Include it in the register RPC so the fingerprint covers it?
2. MetaServer entries and SSD index keys would inherit the final namespace; cross-node discovery then only matches truly compatible peers. Any migration concern for existing SSD/metaserver data?
3. Should the fingerprint be part of the namespace string (observable, greppable in logs) or a separate validated field?

## Related

`pd_connector` still rejects models whose *config* carries MTP fields even when MTP is disabled (`PdConnector does not support MTP layout` assertion keyed off `num_nextn_predict_layers`). Same family of bug as the one the seal contract fixed for the main connector: decisions keyed off config guesses instead of the actual runtime layer set.

PR #346, opened 2026-06-11
fix(core)!: seal instance layer topology from registered layers

xiaguan:
## Problem

Models that ship optional MTP layers in their hf config (e.g. GLM-5.1 with `num_nextn_predict_layers`) break when the user does **not** enable MTP. #335 sized the layer-id space from a connector-side config guess (`num_hidden_layers + num_nextn_predict_layers`), so the engine expected 158 layers while workers only ever registered 156 — every slot space stayed permanently incomplete and the first save/load failed. External drafters (EAGLE/dflash-style) whose layer indices restart at 0 could additionally collide with `extract_layer_index`-anchored ids.

Root cause: the layer-id space is namespace-level data layout, but the connector guessed it from model config instead of using the layer set that actually exists at runtime.

## Design

The layer set **is** the topology. Registration becomes two-phase:

1. **Registering** — each worker declares only the KV cache layers that actually exist on its device (names + geometry). `num_layers` / `layer_ids` are gone from the register RPC (proto fields `reserved`).
2. **Sealed** — when the `world_size`-th device registers, the engine sorts the union of declared layer names into a dense layer-id space and validates that every `(layer, tp_rank)` slot has an owner. An invalid topology rejects the **registration** (startup), not the first save/load.

This removes every special case: MTP on/off, external drafters, hybrid layouts all work because nothing is guessed — `_build_layer_id_space` and the nextn heuristics are deleted outright (net −131 lines).

Preserved/strengthened semantics:
- **MLA replicas** (#336): several devices may own the same slot via `effective_tp_rank=0`; the same layer claimed by two pipeline stages is rejected at seal.
- **Crash early**: save/load/query require a sealed topology; load now errors loudly on a stored block whose slot layout does not match the instance (was a silent `continue` that left layers uninitialized), and rejects negative destination block ids (the load RPC had no server-side check, unlike save).
- A failed seal rejects the offending registration and the instance keeps waiting, so a corrected worker can retry.

One engine instance serves the workers of one node — not new to sealing: CUDA IPC is same-host only, and the session watcher and query-lease consumption already count to `world_size` on a single server.

Follow-up discussion (namespace topology fingerprint, enforcing layout consistency per namespace server-side): #344

## Breaking change

`register_context_batch` loses `num_layers`/`layer_ids`. The existing exact-match client/server version handshake rejects mixed deployments cleanly.

## Tests

- Rust: workspace suite green; the new seal contract tests (`pegaflow-core/src/instance/tests.rs`) exercise the real `register_new_gpu` path — sorted-id derivation, union-of-workers (PP + MTP layer shapes), MLA replica acceptance, cross-pp rejection, missing-owner rejection + retry, geometry mismatch rejection.
- Multi-GPU box (8×): all multi-device seal tests executed (not skipped) and pass; MLA replica integration test (GLM-5.1 config, 158 caches, 2 devices, save on dev0 → load on dev1, byte-identical) passes.
- Python default gate: 140 passed.
- vLLM E2E correctness matrix in progress on the same box (Qwen3-4B TP1/TP2; DeepSeek-V2-Lite TP2 / PP2 / TP2×PP2 / TP2 per-layer mode) — results will be posted here.

Includes a test-helper fix: the e2e harness now strips `CUDA_VISIBLE_DEVICES` for the spawned pegaflow-server, since the connector registers global device ids (a masked pytest run previously failed registration with "pinned to device N but got M").


xiaguan:
## Multi-GPU test results (8-GPU box, consumer Blackwell sm_120)

**Rust seal contract tests** — `cargo test -p pegaflow-core` with multiple CUDA devices: 108 passed. The five multi-device seal tests that self-skip on single-GPU boxes all executed for real: `mla_replica_registration_seals`, `seal_rejects_replicas_across_pipeline_stages`, `seal_derives_layer_space_from_union_of_workers`, `seal_rejects_missing_slot_owner`, `seal_rejects_inconsistent_layer_geometry`.

**MLA replica integration test** — GLM-5.1 config (158 caches), 2 devices registering with `effective_tp_rank=0` → seal → save on dev0 → load on dev1, byte-identical: passed.

**vLLM E2E correctness matrix** (`test_vllm_e2e_correctness.py`, 4 tests per config):

| Config | outputs match | cache metrics | no RPC failures | no load failures | Sealed topology log |
|---|---|---|---|---|---|
| Qwen3-4B TP1 | ✅ | ✅ | ✅ | ✅ | — |
| Qwen3-4B TP2 (per-rank slots) | ✅ | ✅ | ✅ | ✅ | — |
| DSV2-Lite TP2 (MLA replica) | ⚠️ | ✅ | ✅ | ✅ | `num_layers=27, tp_size=1, world_size=2, total_slots=27` |
| DSV2-Lite PP2 | ✅ | ✅ | ✅ | ✅ | `num_layers=27, tp_size=1, world_size=2, total_slots=27` |
| DSV2-Lite TP2×PP2 | ⚠️ | ✅ | ✅ | ✅ | `num_layers=27, tp_size=1, world_size=4, total_slots=27` |
| DSV2-Lite TP2 (`PEGAFLOW_CROSS_LAYER_BLOCKS=0`) | ⚠️ | ✅ | ✅ | ✅ | `num_layers=27, tp_size=1, world_size=2, total_slots=27` |

The seal logs show exactly the intended behavior: MLA collapses to `tp_size=1`, PP stages contribute disjoint layer ranges whose union forms the 27-layer id space, and TP2×PP2 waits for all 4 workers before sealing.

⚠️ = `test_execution_plan_outputs_match_baseline` only. On this GPU generation FLASH_ATTN cannot serve MLA, so these runs used `VLLM_TEST_ATTN_BACKEND=TRITON_MLA` (override added in 9f2194d); its batch invariance under TP is imperfect — warm outputs diverge from baseline ~200+ chars into generation with semantically equivalent text. Evidence this is sampler nondeterminism rather than KV corruption: the PP2 run (same save/load path, no TP) matches baseline exactly, and corrupted KV would diverge from the first token. The TP2 data-path tests (RPC health, load failures, cache metrics) are all green.

Also fixed while setting this up: the e2e harness inherited the pytest `CUDA_VISIBLE_DEVICES` mask into pegaflow-server, which broke the connector's global-device-id contract (e5f0725).

feifei-111 (approved):

PR #349, opened 2026-06-12
feat(server): add p2p RDMA fetch benchmark example

xiaguan:
First batch split out of #342: the standalone `p2p_bench` example, adapted to the current master API.

## What

`pegaflow-server/examples/p2p_bench.rs` — an end-to-end p2p RDMA fetch benchmark and integrity test that drives the production cross-node path without vLLM:

- **Holder**: saves blocks through the real GPU save path with a vLLM-shaped layout (layers × tp_ranks slots, split K/V segments), registers them in an in-process MetaServer, and serves the production Engine gRPC service.
- **Requester**: registers the same layout, then walks the real prefetch path: MetaServer discovery → `QueryBlocksForTransfer` → RDMA READ → SealedBlock rebuild. With `--verify` it loads every fetched block back to GPU and checks every byte against the deterministic generator pattern.

## Changes vs the #342 version

- Updated `register_context_layer_batch` call to the post-#340 signature (layer ids and layer count are no longer caller-supplied).
- Doc comment describes the current RDMA READ fetch flow instead of the v2 PushBlocks flow; when the v2 engine lands the example body stays the same.
- Cargo.lock: sync `pegaflow-pd-wire` to 0.22.9 (was stale after the version bump).

No Cargo.toml changes needed — `clap`/`cudarc` are existing deps and `pegaflow-metaserver` is already a dev-dependency.

## Verification

- `cargo build --no-default-features --features cuda-13,rdma -p pegaflow-server --example p2p_bench` passes.
- fmt / clippy / typos / cargo test (prek hooks) pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-pd-wire"
-version = "0.22.8"
+version = "0.22.9"
 dependencies = [
  "serde",
  "serde_json",
--- a/pegaflow-core/src/backing/rdma_fetch.rs
+++ b/pegaflow-core/src/backing/rdma_fetch.rs
@@ -592,7 +592,13 @@ fn get_or_create_channel(
         .map_err(|e| format!("invalid remote address: {e}"))?
         .connect_timeout(Duration::from_secs(5))
         .connect_lazy();
-    let client = EngineClient::new(channel);
+    // Match the engine server's 64 MiB message cap: a QueryBlocksForTransfer
+    // response carries per-slot transfer descriptors, so a large block batch
+    // overflows tonic's default 4 MiB decode limit.
+    const MAX_GRPC_MESSAGE_SIZE: usize = 64 * 1024 * 1024;
+    let client = EngineClient::new(channel)
+        .max_decoding_message_size(MAX_GRPC_MESSAGE_SIZE)
+        .max_encoding_message_size(MAX_GRPC_MESSAGE_SIZE);
     cache.insert(addr.to_string(), client.clone());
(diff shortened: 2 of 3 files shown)


xiaguan:
Ran the example end-to-end on two 8×H200 nodes (8× 400G IB NICs each, holder + requester):

**Default shape** (`layers=36 tp=8 kv_bytes=4096 blocks=1000`, 2250 MiB/set — realistic dense-model KV block size, 576k 4 KiB segments per set):
```
FETCH measure set=1 mib=2250.0 ms=347.35 gib_s=6.33
FETCH measure set=2 mib=2250.0 ms=365.40 gib_s=6.01
FETCH measure set=3 mib=2250.0 ms=354.86 gib_s=6.19
SUMMARY sets_measured=3 set_mib=2250.0 median_ms=354.86 median_gib_s=6.19
```

**Bandwidth shape** (`kv_bytes=65536 blocks=200`, 7200 MiB/set):
```
FETCH measure set=1 mib=7200.0 ms=200.58 gib_s=35.05
FETCH measure set=2 mib=7200.0 ms=158.54 gib_s=44.35
SUMMARY sets_measured=2 set_mib=7200.0 median_ms=200.58 median_gib_s=35.05
```

`--verify` passed on every set in both runs (byte-exact across all 8 ranks).

The default shape is what exposed the 4 MiB tonic decode limit fixed in 50c8b64 — the first run failed every fetch with `decoded message length too large: found 7213044 bytes`. The small-segment shape is IOPS/descriptor-bound (~6 GiB/s); the v2 push engine (#342) is the follow-up that targets exactly this path.

jimmy-evo (approved):

PR #354, opened 2026-06-16
feat(transfer)!: encode per-slot schema + NUMA source map in QueryBlocksForTransfer

xiaguan:
## What & why

`QueryBlocksForTransfer` previously returned per-block / per-slot RDMA metadata as nested protobuf (`TransferBlockInfo` / `TransferSlotInfo`), with the wire size scaling as `blocks × slots`. This branch replaces that with a compact, run-encoded **`TransferPlan`** (new `pegaflow-transfer-wire` crate, postcard-encoded into a single `bytes transfer_plan` field): the response carries a per-slot **segment schema** plus a per-slot **NUMA source map** once, and the requester rebuilds each block's slabs on the NUMA the holder actually used.

The "schema source map" point: the holder tells the requester not just *where* the bytes live, but *which NUMA* each slot is on, so the cross-node RDMA fetch materializes and rebuilds NUMA-correctly instead of landing everything on one node.

## What's in it (5 commits)

- **feat: TransferPlan wire encoding** — new `pegaflow-transfer-wire` crate + `transfer_plan/{encode,schema}`; `QueryBlocksForTransferResponse` carries `bytes transfer_plan` (postcard) instead of nested messages. Wire size scales with *blocks*, not *blocks × slots*.
- **fix: split oversized RDMA reads at 128MB** + **fix: enable IBV_ACCESS_RELAXED_ORDERING** — RDMA read-path fixes the plan-driven fetch depends on.
- **feat: per-slot NUMA source map** — thread holder-side per-slot NUMA through `transfer_plan/{view,merge,schema,rebuild}` + `materialize`; the requester allocates one slab per distinct chunk NUMA and rebuilds each slot on its source NUMA. Rebuild **rejects UNKNOWN NUMA** (crash-early; encode warns and the fetch boundary logs + meters the error, so the degradation is never silent). Restores `SealedBlock::from_slots` to build with empty `slot_numas` (SSD-prefetch path; was regressed into a guaranteed panic). Adds the p2p RDMA integration coverage.
- **test: assert rebuild rejects unknown NUMA loudly** — locks in the crash-early guard so it can't silently regress.

Also: the kernel-backend integration test now self-skips when the driver can't JIT-load its PTX (older driver vs newer CUDA toolkit) instead of failing the suite — it still runs wherever the driver supports the kernel.

## Breaking change

⚠️ Wire-incompatible: `QueryBlocksForTransferResponse` fields are renumbered and `TransferBlockInfo` / `TransferSlotInfo` are removed in favor of `bytes transfer_plan`. **Holder and requester must be upgraded together.**

## Testing

- Local gate green (cuda-13,rdma): `cargo fmt` / `cargo clippy -D warnings` / `cargo test --release` all pass.
- Unit coverage in `transfer_plan::tests`: per-block NUMA preservation through rebuild, one slab per distinct chunk NUMA, loud UNKNOWN-NUMA rejection, encode→materialize→rebuild byte round-trip, wire-size scaling.
- **p2p RDMA integration tests run on an 8-GPU, 2-NUMA-socket node with 400G InfiniBand** — all 4 pass, byte-exact:
  - normal MHA TP>1 → cross-slot multi-NUMA exercised (`distinct={0,1}`)
  - MLA replica (no DCP) → cross-block multi-NUMA via server save-NUMA round-robin (`distinct={0,1}`)
  - single-writer baseline
  - The multi-NUMA cases self-adapt to single-NUMA boxes (skip the cross-NUMA assertion, still verify topology + data roundtrip).

## Note

Related transfer work is in #350 (run-encode `QueryBlocksForTransfer` addresses) on a separate branch — no shared commits, but both touch the transfer fetch path, so mind merge order.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1927,6 +1927,7 @@ dependencies = [
  "pegaflow-common",
  "pegaflow-proto",
  "pegaflow-transfer",
+ "pegaflow-transfer-wire",
  "rand 0.10.0",
  "shared_memory",
  "smallvec",
@@ -2050,6 +2051,15 @@ dependencies = [
  "syscalls",
  "thiserror 2.0.18",
  "tokio",
+]
+
+[[package]]
+name = "pegaflow-transfer-wire"
+version = "0.22.9"
+dependencies = [
+ "postcard",
+ "serde",
+ "smallvec",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2012,6 +2012,7 @@ dependencies = [
  "pegaflow-core",
  "pegaflow-metaserver",
  "pegaflow-proto",
+ "pegaflow-transfer-wire",
  "prometheus",
  "pyo3",
  "pyo3-build-config",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
   "pegaflow-server",
   "pegaflow-metaserver",
   "pegaflow-pd-wire",
+  "pegaflow-transfer-wire",
   "pegaflow-transfer",
   "python",
 ]
@@ -21,6 +22,7 @@ license = "Apache-2.0"
 pegaflow-common = { path = "pegaflow-common" }
 pegaflow-core = { path = "pegaflow-core", default-features = false }
 pegaflow-pd-wire = { path = "pegaflow-pd-wire" }
+pegaflow-transfer-wire = { path = "pegaflow-transfer-wire" }
 pegaflow-transfer = { path = "pegaflow-transfer", default-features = false }
 sideway = "0.4.3"
 rdma-mummy-sys = "0.2.4"
--- a/pegaflow-core/Cargo.toml
+++ b/pegaflow-core/Cargo.toml
@@ -36,6 +36,7 @@ tonic.workspace = true
 dashmap.workspace = true
 smallvec.workspace = true
 pegaflow-proto = { path = "../pegaflow-proto" }
+pegaflow-transfer-wire.workspace = true
 pegaflow-transfer = { path = "../pegaflow-transfer", default-features = false, optional = true }
 mea.workspace = true
 
--- a/pegaflow-core/Cargo.toml
+++ b/pegaflow-core/Cargo.toml
@@ -12,6 +12,8 @@ cuda-12 = ["cudarc/cuda-12080", "cudarc/nvrtc", "pegaflow-transfer?/cuda-12"]
 cuda-13 = ["cudarc/cuda-13000", "cudarc/nvrtc", "pegaflow-transfer?/cuda-13"]
 rdma = ["dep:pegaflow-transfer"]
 tracing = ["fastrace"]
+# Exposes `bench_support` for criterion benches; not for production builds.
+bench = []
 
 [dependencies]
 pegaflow-common.workspace = true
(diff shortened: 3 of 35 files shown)


xiaguan:
## Perf: large-segment NIC-striping fix (cedfc25)

While validating this branch's p2p RDMA fetch path I found (and fixed in `cedfc25`) a **large-segment throughput regression vs master**.

**Root cause.** The coalesced `TransferPlan` hands the transfer backend one big `TransferDesc` per NUMA node. `batch_transfer_async` mapped each desc to a single NIC (and, via the per-desc QP bucket, a single QP), so a multi-GB read saturated one NIC while the rest sat idle — aggregate throughput capped at roughly one NUMA group's worth of NICs.

**Fix.** Stripe each desc into 8 MiB sub-ranges, round-robin across its NUMA node's NICs (and their QPs), so one large coalesced read fans out over the whole group. The upper layer stays agnostic — it submits byte ranges; fanning them across the fabric is the transfer library's job.

**Measured** — single node, 8×400G IB NICs, 2 NUMA sockets, NIC loopback, ~10 GiB/set, median of 3 measured sets (box under other load, so absolute numbers are loopback+contention-bound; read the **relative** A/B/C):

| segment | master | this branch (pre-fix) | this branch (+fix) |
|---|---:|---:|---:|
| 4 KiB   | 7.14   | 10.33 | **15.72** |
| 64 KiB  | 66.01  | 25.24 | **95.05** |
| 256 KiB | 104.67 | 26.75 | **123.33** |

Pre-fix the branch fell to **0.26–0.38×** of master at large segments; with the fix it's **1.2–2.2×** master across the board — coalescing cuts WQE overhead at the IOPS end, striping restores full NIC parallelism at the bandwidth end.

_Caveat: single-node NIC loopback is not wire bandwidth; numbers are for relative comparison only. A unit test (`large_desc_fans_out_across_every_nic_in_group`) locks in the fan-out invariant without hardware._

xiaguan:
## Small-batch regression check

The Performance numbers in the description are large-batch. This confirms tiny fetches (1-block warm hit, MLA single-block read) pay **no** fixed-overhead penalty from the TransferPlan encode / geometry / postcard work. Same 8-GPU, 2-NUMA, 8×400G node, single-host loopback, warm measure set (median of 4–5), byte-verified on all 8 ranks.

| blocks | bytes/set | master e2e | this branch e2e | |
|---|---|---|---|---|
| 1 | 2.2 MiB | 2.03 ms | **1.59 ms** | 1.3× |
| 8 | 18 MiB | 3.41 ms | **2.20 ms** | 1.6× |
| 64 | 144 MiB | 16.38 ms | **5.87 ms** | 2.8× |

**No regression — faster at every size.** Even at 1 block the branch wins: master still emits per-block RDMA descriptors (1 block = 576 cells across 36 layers × TP8 × split-K/V) plus nested protobuf, whereas the TransferPlan coalesces to 2 descriptors, so `build_transfer_tasks` + `submit_transfer` stay ~0 on the branch vs master's 0.1 ms (1 blk) → ~7 ms (64 blk), and `query` is cheaper too. The SoA rebuild is serial, so there is no thread-spawn cost on small fetches — the 1-block fixed overhead is just `query ~0.15 ms` + `rebuild ~0.02 ms`.

Stage detail at the extremes (ms, measure set, connect amortized to 0):

| | query | build+submit | rdma | rebuild |
|---|---|---|---|---|
| 1 blk master | 0.20 | 0.16 | 0.04 | 0.03 |
| 1 blk branch | 0.15 | ~0.01 | 0.10 | 0.02 |
| 64 blk master | 5.5 | ~7.0 | 1.75 | 1.4 |
| 64 blk branch | 0.95 | ~0.07 | 1.0 | 1.4 |

The only line item where the branch is ever slightly higher is `rdma_wait` at ≤8 blocks (sub-0.1 ms of coalesced-read setup), and it is already lower by 64 blocks.

PR #355, opened 2026-06-16
refactor(pd): restructure pd_connector for maintainability

GentleCold:
## What

Maintainability refactor of `python/pegaflow/pd_connector/` (the vLLM P/D RDMA-push connector). Behavior-preserving throughout — no change to the vLLM-facing API, the RDMA data flow, or the wire format.

## Why

The package had grown to ~6.4k lines with parallel-but-unshared structure: three near-identical connector facades, duplicated async-executor scaffolding in three places, a 1.2k-line `prefill_worker.py`, and two flat clusters of decode state with transitions scattered across the handler.

## Changes (6 commits, each independently green against the unit gate)

1. **Shared async task runner** — factor the duplicated `ThreadPoolExecutor` + condition/inflight skeleton out of the prefill push sender, push finalizer, and decode RDMA-done waiter into `async_runner.py` (`AsyncTaskPool` / `InflightTaskRunner`). Per-request counting, cancellation, and dedup stay in subclasses where their semantics differ.
2. **Collapse connector facades** — move the ~11 identical vLLM callbacks shared by `PdDecodeConnector` / `PdPrefillConnector` into a `_PdSplitConnector` base; each side overrides only what differs.
3. **Split `prefill_worker.py`** — into `prefill_tasks.py` (data carriers), `prefill_async.py` (executors + RDMA-stat helpers), and `prefill_worker.py` (handler). Moved symbols are re-imported so they still resolve as module globals.
4. **Group metadata codecs** — reorder the wire (de)serialization helpers by type so each encode/decode pair sits together. Pure reordering, no byte-format change.
5. **Encapsulate decode wait/failure state** — the 7 request/failure collections + lock move into `_DecodeWaitState` with named transitions and one drain method per consuming vLLM callback. The three failure channels stay separate (they are drained by different callbacks); this is encapsulation, not a merge.
6. **Extract decode peer topology + IMM allocation** — the peer-layout / MR-desc / layer-template / expected-imm-count / imm-id cluster and its 6 methods move into `_DecodePeerState`; `_peer_block_size` becomes a derived property.

## Testing

- **Unit gate**: `cd python && uv run --extra test pytest tests/pd/test_pd_connector_flow.py tests/pd/test_pd_connector_layout.py` — 117 passed after every commit.
- **Real RDMA bench** (single-node H200, Qwen3-8B, p-tp4/d-tp4, in 8192 / out 128, c8, 128 prompts):
  - PD (this branch): 128/0 completed, mean TTFT 246 ms, mean TPOT 3.40 ms, 95.9k tok/s — decode logs show 1729 RDMA done / prefill 1152 push done, **0 errors**.
  - On par with NIXL (TTFT 244 ms, 97.7k tok/s) and baseline 2×TP4-router; correctness matches NIXL (token-exact on short greedy decode).
- Strict RDMA NIC + CPU pinning preserved throughout.

## Notes

No behavior or performance regression.

### DIFF
--- a/python/pegaflow/pd_connector/__init__.py
+++ b/python/pegaflow/pd_connector/__init__.py
@@ -26,29 +26,36 @@ from pegaflow.pd_connector.worker import (
 )
 
 
-class PdDecodeConnector(PdConnectorClassMixin, KVConnectorBase_V1, SupportsHMA):
-    """Decode-side vLLM connector for P/D RDMA push."""
+class _PdSplitConnector(PdConnectorClassMixin, KVConnectorBase_V1, SupportsHMA):
+    """Base for the split decode/prefill connectors.
 
-    @classmethod
-    def requires_piecewise_for_cudagraph(cls, extra_config: dict[str, Any]) -> bool:
-        return False
+    Holds the vLLM callbacks that are identical on both sides. Subclasses set
+    ``_scheduler_cls`` / ``_worker_cls`` and override only the callbacks that
+    actually differ between decode and prefill (load/save side, cudagraph
+    requirement, load-error reporting).
+    """
+
+    _scheduler_cls: type
+    _worker_cls: type
 
     def __init__(self, vllm_config: Any, role: KVConnectorRole, kv_cache_config: Any = None):
         super().__init__(vllm_config, role, kv_cache_config)
         assert_supported_config(vllm_config)
-        self._scheduler: PdDecodeSchedulerConnector | None = None
-        self._worker: PdDecodeWorkerConnector | None = None
+        self._scheduler: Any | None = None
+        self._worker: Any | None = None
         self._metrics = PdMetricsTracker()
         if role == KVConnectorRole.SCHEDULER:
-            self._scheduler = PdDecodeSchedulerConnector(vllm_config, metrics=self._metrics)
+            self._scheduler = self._scheduler_cls(vllm_config, metrics=self._metrics)
         elif role == KVConnectorRole.WORKER:
-            self._worker = PdDecodeWorkerConnector(
+            self._worker = self._worker_cls(
                 vllm_config,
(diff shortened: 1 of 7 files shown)


GentleCold:
Thanks for the review. Both findings are accurate; addressed differently per scope:

**#1 `has_wait` lock (decode_worker.py:62)** — Fixed in fed9aa7. `_DecodeWaitState` documents itself as lock-guarded, so `has_wait()` now acquires the lock for its membership read, consistent with the concurrent waiter/prefill callbacks. (The call site is the single-threaded forward path, so this is hardening rather than a live race, but the docstring should not over-claim.)

**#2 `build_all_rank_handshake_dicts` KeyError in all_gather fallback (decode_worker.py:266)** — Correct, but this is **pre-existing behavior** (identical at the base commit before this refactor) and a behavior change to fix, so it's intentionally out of scope for this behavior-preserving refactor. It is only reachable on the degraded path where `torch.distributed.all_gather_object` raises on a multi-rank (`tp_size > 1`) decode worker; even then `build_all_rank_handshake_dicts` is only called via `_dispatch_prefill`, which is gated by `tp_rank == 0`. Tracking it as a separate follow-up rather than mixing a logic fix into the restructure.

GentleCold:
Filed #2 as a tracked follow-up: novitalabs/pegaflow#356.

GentleCold:
#2 (build_all_rank_handshake_dicts KeyError on all_gather fallback) is now fixed in 05ab7a9 — iterating `sorted(layer_templates)` (the ranks actually gathered) instead of `range(tp_size)`. Normal gathered path is unchanged (layer_templates holds every rank 0..tp_size-1); the fallback now builds only the local rank's handshake, matching the warning. Closing #356 as resolved here.

xiaguan (approved):

PR #359, opened 2026-06-17
feat(connector): distribute MLA TP save across ranks by layer

xiaguan:
## Problem

MLA replicates the latent KV cache across every TP rank (it is not sharded by head like MHA). On the save path PegaFlow collapsed that to a single writer: only TP0 submitted saves, so the **entire model's KV was streamed to CPU through one GPU's copy engine and one PCIe link**, while the other ranks sat idle. A round-robin NUMA hint then scattered half of TP0's writes to a remote NUMA node, adding cross-socket traffic.

The read path was already N-way parallel (every rank H2D-loads its own full replica), so only the write path had a single-card bottleneck.

## Change

Treat MLA's replicated TP like PP for saving: **split the layer set across ranks**.

- Each rank computes a disjoint layer slice by `sorted_layer_name index % tp_size == tp_rank` and saves only that slice, reading from **its own** GPU into its **NUMA-local** pinned pool. Sorting by name aligns with how the server seals topology, so the partition is exact (no gaps, no double-writes).
- All ranks now run the full save lifecycle (empty trailing slices included) so vLLM's per-rank `finished_sending` aggregation stays correct.
- Load is unchanged — each rank still reads its full replica.
- Stop **silently** downgrading cross-layer blocks for MLA: MLA attention backends don't expose the per-layer stride needed for `use_uniform_kv_cache`, so cross-layer never actually built. We now log it instead of pretending it worked.
- Remove the obsolete round-robin save NUMA hint (and its server/engine plumbing); save now always lands NUMA-local. Net: **+144 / −359**.

Sharding only engages on the MLA-replica path (`is_mla && !mla_layer_split_kv_cache && dcp_world_size == 1 && tp_size > 1`). Non-MLA ranks and the existing `mla_layer_split_kv_cache` feature already store disjoint sets and are untouched.

## Validation

End-to-end on an 8-GPU host with DeepSeek-V2-Lite (27 layers), reading the engine's per-device save/load debug logs:

| Topology | SAVE (layers per device) | LOAD (layers per device) |
| --- | --- | --- |
| TP2 | `14, 13` (Σ27) | `27, 27` |
| PP2 | `14, 13` (Σ27, PP-native; sharding off) | `14, 13` |
| TP2×PP2 | `7, 7, 7, 6` (Σ27) | `14, 14, 13, 13` |
| TP8 | `4,4,4, 3,3,3,3,3` (Σ27) | `27 ×8` |

Save distributes a disjoint 1/N slice across every participating GPU (summing to the full layer set); load stays full-replica within each TP group. `PP2` confirms the non-shard path is a clean no-op. The `cache_metrics`, `no_data_path_rpc_failures`, and `no_kv_load_failures` e2e assertions pass for every topology. The `outputs_match_baseline` assertion flakes under TP only — a known batch-invariance limitation of the MLA attention backend required on this GPU generation (the PP-only run passes it), unrelated to this change.

## Out of scope

- MLA + DCP (`dcp_world_size > 1`): each DCP rank already stores different tokens.
- Localizing the read-side N× fan-out (e.g. NUMA-aware storage redundancy for loads).

### DIFF
--- a/pegaflow-core/src/instance.rs
+++ b/pegaflow-core/src/instance.rs
@@ -529,56 +529,9 @@ impl InstanceContext {
         &self.namespace
     }
 
-    /// Access the effective TP size registered with the engine.
-    pub(crate) fn tp_size(&self) -> usize {
-        self.tp_size
-    }
-
     /// Total worker count registered for this instance.
     pub(crate) fn world_size(&self) -> usize {
         self.world_size
-    }
-
-    /// Return unique valid NUMA nodes for registered shards in one save group.
-    pub(crate) fn registered_numa_nodes_for_save_group(
-        &self,
-        tp_rank: usize,
-        pp_rank: usize,
-    ) -> Vec<NumaNode> {
-        let state = self.state.lock();
-        let mut nodes: Vec<NumaNode> = state
-            .gpu_contexts
-            .values()
-            .filter(|gpu| gpu.tp_rank() == tp_rank && gpu.pp_rank() == pp_rank)
-            .map(|gpu| gpu.preferred_numa())
-            .filter(|node| node.is_valid())
-            .collect();
-        nodes.sort_unstable();
-        nodes.dedup();
-        nodes
-    }
-
-    /// Validate that a save placement hint targets a registered shard NUMA node.
-    pub(crate) fn validate_save_numa_hint(
-        &self,
-        tp_rank: usize,
-        pp_rank: usize,
-        numa_node: NumaNode,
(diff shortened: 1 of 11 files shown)


xiaguan:
CI result for head 2bcec9a: PASS

Checks run:
- GLM-5.1 performance CI: PASS
- Kimi-K2.6 performance CI: PASS
- GLM-5.1 accuracy CI: PASS
- Kimi-K2.6 accuracy CI: PASS

feifei-111 (approved):

PR #369, opened 2026-06-26
fix(metaserver): size removal pipeline for eviction bursts

xiaguan:
## Problem

A pegaflow-server deployment was dropping **~56% of its MetaServer block-removal notifications** (44–62% per node, three nodes over ~4.7h). Measured from Prometheus counters, with an exact accounting identity holding on every node:

```
cache_block_evictions == metaserver_removal_blocks (sent) + metaserver_removal_queue_full (dropped)
```

Effect: the MetaServer owner directory carries stale owners, so a peer that queries it gets routed to a node that already evicted the block → wasted cross-node RDMA prefetch (`blocks=0/N`). Correctness is fine (TTL sweep is the fallback), but it's a steady efficiency leak, not an occasional one.

## Mechanism

LRU eviction runs as a synchronous burst inside `StorageEngine::reclaim_until_allocator_can_allocate`: each iteration evicts `RECLAIM_BATCH_SIZE` blocks and fires **one** fire-and-forget removal command into a shared mpsc channel of depth `DEFAULT_METASERVER_QUEUE_DEPTH`. The single async consumer drains + coalesces but does one MetaServer RPC at a time. Page-first save allocates **one page per block in a loop** (`offload.rs:205`), so a long request is hundreds/thousands of `allocate()` calls, each able to trigger a reclaim batch — the producer easily overruns the channel during a burst and `try_send` drops the tail.

This is **not** memory pressure (`pool_alloc_failures = 0`) and **not** insert/remove starvation (`registration_queue_full = 0`, inserts never drop). It is purely removal burstiness.

## Change

- **`RECLAIM_BATCH_SIZE` 64 → 512** (load-bearing): cuts removal command rate ~8×. Over-eviction is bounded by one batch (~512 pages ≈ ~0.15% of a multi-hundred-GB pinned pool); eviction only unlinks LRU entries, in-transfer blocks stay alive via their `Arc`.
- **`DEFAULT_METASERVER_QUEUE_DEPTH` 256 → 4096**: absorbs residual bursts. Also the default for the `--metaserver-queue-depth` CLI flag (stale help text fixed in the same change).

## Trade-offs (honest)

- A reclaim batch now holds the `read_cache` mutex ~8× longer per call — reclaim turns from frequent small pauses into rarer, heavier ones; concurrent prefix lookups/inserts can see a deeper tail-latency spike when a reclaim lands.
- `allocate()` tail latency is concentrated, not removed: 511/512 calls still return on hit, the 1/512 that reclaims pays for 512 evictions at once. Per-request total reclaim cost is unchanged.
- Worst-case channel memory ~100–200 MB when pathologically full (noise vs the pool); near-zero at steady state since the consumer drains the whole channel per wake.

## Validation

No deterministic test — the contract (burst-time `sent + dropped == evictions`, drop → ~0) is async/RPC-timing dependent and would only be reproducible with heavy mocks. Confidence comes from the production drop measurement above. **Merge gate: re-measure `metaserver_removal_queue_full` → ~0 after rollout.**

### DIFF
--- a/pegaflow-core/src/internode/metaserver_client.rs
+++ b/pegaflow-core/src/internode/metaserver_client.rs
@@ -17,7 +17,9 @@ use uuid::Uuid;
 
 use crate::metrics::core_metrics;
 
-pub const DEFAULT_METASERVER_QUEUE_DEPTH: usize = 256;
+// Shared insert/remove command channel depth. Eviction bursts outrun the single
+// consumer's per-RPC drain, so a shallow queue silently drops removals.
+pub const DEFAULT_METASERVER_QUEUE_DEPTH: usize = 4096;
 
 /// Error type for MetaServer client operations.
 #[cfg(feature = "rdma")]
--- a/pegaflow-core/src/internode/metaserver_client.rs
+++ b/pegaflow-core/src/internode/metaserver_client.rs
@@ -416,10 +416,10 @@ async fn registration_loop(
 
         // Process inserts
         let insert_namespaces: Vec<(String, Vec<Vec<u8>>)> = inserts.into_iter().collect();
-        let mut insert_failed_at = None;
+        let mut insert_failed_at: Option<(usize, usize)> = None;
 
         'insert: for (i, (namespace, hashes)) in insert_namespaces.iter().enumerate() {
-            for chunk in hashes.chunks(MAX_HASHES_PER_RPC) {
+            for (chunk_idx, chunk) in hashes.chunks(MAX_HASHES_PER_RPC).enumerate() {
                 let count = chunk.len();
                 let request = InsertBlockHashesRequest {
                     namespace: namespace.clone(),
@@ -449,15 +449,15 @@ async fn registration_loop(
                             core_metrics().metaserver_session_resets.add(1, &[]);
                             heartbeat.node_registered = false;
                         }
-                        insert_failed_at = Some(i);
+                        insert_failed_at = Some((i, chunk_idx * MAX_HASHES_PER_RPC));
                         break 'insert;
                     }
                 }
             }
         }
 
-        if let Some(idx) = insert_failed_at {
-            let dropped: usize = insert_namespaces[idx..].iter().map(|(_, h)| h.len()).sum();
+        if let Some((idx, offset)) = insert_failed_at {
+            let dropped = unsent_after_failure(&insert_namespaces, idx, offset);
             core_metrics()
                 .metaserver_registration_failures
                 .add(dropped as u64, &[]);
@@ -473,10 +473,10 @@ async fn registration_loop(
 
         // Process removes
         let remove_namespaces: Vec<(String, Vec<Vec<u8>>)> = removes.into_iter().collect();
-        let mut remove_failed_at = None;
+        let mut remove_failed_at: Option<(usize, usize)> = None;
 
         'remove: for (i, (namespace, hashes)) in remove_namespaces.iter().enumerate() {
-            for chunk in hashes.chunks(MAX_HASHES_PER_RPC) {
+            for (chunk_idx, chunk) in hashes.chunks(MAX_HASHES_PER_RPC).enumerate() {
                 let count = chunk.len();
                 let request = RemoveBlockHashesRequest {
                     namespace: namespace.clone(),
@@ -506,15 +506,15 @@ async fn registration_loop(
                             core_metrics().metaserver_session_resets.add(1, &[]);
                             heartbeat.node_registered = false;
                         }
-                        remove_failed_at = Some(i);
+                        remove_failed_at = Some((i, chunk_idx * MAX_HASHES_PER_RPC));
                         break 'remove;
                     }
                 }
             }
         }
 
-        if let Some(idx) = remove_failed_at {
-            let dropped: usize = remove_namespaces[idx..].iter().map(|(_, h)| h.len()).sum();
+        if let Some((idx, offset)) = remove_failed_at {
+            let dropped = unsent_after_failure(&remove_namespaces, idx, offset);
             core_metrics()
                 .metaserver_removal_failures
                 .add(dropped as u64, &[]);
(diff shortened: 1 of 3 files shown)


GentleCold (commented):
Thanks for the detailed production analysis. The direction makes sense to me: increasing `RECLAIM_BATCH_SIZE` reduces the number of removal commands produced during reclaim bursts, and increasing the shared MetaServer command queue gives the async consumer more room to absorb short spikes.

One concern before merging: this also increases the worst-case size of a single MetaServer remove RPC. The consumer drains the mpsc queue and coalesces removals by namespace before sending `RemoveBlockHashesRequest`. With the new defaults, one drain can theoretically contain:

`4096 queue entries * 512 hashes = 2,097,152 hashes`

In the normal vLLM path, `Request.block_hashes` uses the default `sha256` prefix-cache hash, so each hash is 32 bytes. That makes the raw hash payload about 64 MiB, and the protobuf `repeated bytes` payload about 68 MiB before adding the small `namespace`, `node`, and `node_id` fields. MetaServer currently constructs `MetaServerServer::new(service)` without setting `max_decoding_message_size`, so this can exceed tonic’s default 4 MiB receive limit and turn queue drops into `remove_block_hashes` RPC failures under a large burst.

I think the queue/batch sizing is reasonable, but the consumer should cap RPC batch size explicitly. For example, after draining/coalescing, split each namespace’s hashes into fixed-size chunks before calling `remove_block_hashes` / `insert_block_hashes`. That keeps the queue useful as a burst buffer without letting RPC size become `queue_depth * reclaim_batch_size`.

Suggested validation after the change:
- `pegaflow_metaserver_removal_queue_full` drops to ~0
- `pegaflow_metaserver_removal_failures` does not increase
- MetaServer `remove_block_hashes` RPC latency remains acceptable
- stale-owner symptoms such as remote prefetch `blocks=0/N` decrease

xiaguan:
Thanks @GentleCold — you're right, and I confirmed both halves in the code: the consumer coalesces a whole queue drain per namespace into a single `RemoveBlockHashesRequest`/`InsertBlockHashesRequest` with no size cap (`registration_loop`), and `pegaflow-metaserver/src/lib.rs` builds `MetaServerServer::new(service)` without `max_decoding_message_size`, so it's on tonic's 4 MiB default. With the new sizing, worst-case RPC = `queue_depth * batch_size` hashes, which blows past that and turns queue drops into `remove_block_hashes` failures. A single-namespace server (the common case) doesn't even get the namespace-spreading relief.

Fixed in 1bc2dc2 by capping the RPC batch on the consumer side, exactly as you suggested:

- New `MAX_HASHES_PER_RPC = 16384`; after coalescing, each namespace's hashes are sent in `.chunks(MAX_HASHES_PER_RPC)`. A full chunk of 32-byte sha256 hashes is ~0.5 MiB, well under the 4 MiB limit, and RPC size is now decoupled from `queue_depth`/`reclaim_batch_size`. Applied to both the insert and remove paths.
- Deterministic test `large_namespace_removal_splits_into_bounded_rpcs`: drives a single-namespace removal of `MAX_HASHES_PER_RPC * 10 + 1` (~5 MiB raw, over the 4 MiB default) through an in-process MetaServer and asserts it splits into 11 bounded chunks (each <= cap) with no hashes lost or duplicated. Without chunking that single RPC would be rejected by the default decode limit, so the test genuinely exercises the boundary.

Your four validation items stay as the rollout merge gate (`removal_queue_full` -> ~0, `removal_failures` flat, `remove_block_hashes` latency OK, `blocks=0/N` down) — I'll capture those when this lands on the deployment.

GentleCold:
Thanks for the chunking fix. One follow-up on the failure accounting: after splitting a namespace into `MAX_HASHES_PER_RPC` chunks, the current `insert_failed_at` / `remove_failed_at` still only records the namespace index. If a later chunk in that namespace fails after earlier chunks were already sent successfully, the failure counter sums the whole namespace again:

```rust
let dropped: usize = remove_namespaces[idx..].iter().map(|(_, h)| h.len()).sum();
```

That over-counts `metaserver_registration_failures` / `metaserver_removal_failures` by including chunks from the failed namespace that already succeeded. This matters because `removal_failures` is part of the rollout gate.

Suggested shape: track the failed chunk start offset inside the namespace, then count only the failed chunk and the remaining unsent hashes in that namespace, plus all later namespaces. For example:

```rust
let mut remove_failed_at: Option<(usize, usize)> = None; // (namespace_idx, hash_offset)

'remove: for (i, (namespace, hashes)) in remove_namespaces.iter().enumerate() {
    for (chunk_idx, chunk) in hashes.chunks(MAX_HASHES_PER_RPC).enumerate() {
        // ... send RPC ...
        if let Err(e) = c.remove_block_hashes(request).await {
            let failed_offset = chunk_idx * MAX_HASHES_PER_RPC;
            remove_failed_at = Some((i, failed_offset));
            break 'remove;
        }
    }
}

if let Some((idx, offset)) = remove_failed_at {
    let current_ns_remaining = remove_namespaces[idx].1.len().saturating_sub(offset);
    let later_namespaces: usize = remove_namespaces[idx + 1..]
        .iter()
        .map(|(_, h)| h.len())
        .sum();
    core_metrics()
        .metaserver_removal_failures
        .add((current_ns_remaining + later_namespaces) as u64, &[]);
    client = None;
}
```

Same issue applies to the insert path.

xiaguan:
Good catch — fixed in ebb3a74. You're right that re-counting the whole namespace inflates the failure metric, and it matters precisely because `removal_failures` is one of the rollout gates.

`insert_failed_at` / `remove_failed_at` are now `Option<(usize, usize)>` = `(namespace_idx, hash_offset)`, where `hash_offset = chunk_idx * MAX_HASHES_PER_RPC` at the failing chunk. The dropped count is the unsent tail of that namespace (`len - offset`, earlier chunks already landed) plus all later namespaces — extracted into a pure `unsent_after_failure` helper so the insert and remove paths share one implementation.

Added a unit test `unsent_after_failure_counts_only_unsent_tail` covering: first-chunk failure (whole namespace counts), mid-namespace failure (only the tail), last-namespace failure (no later namespaces), and offset past the namespace length (saturates to 0). I tested the helper rather than the metric itself because the OTLP counters are process-global and not unit-assertable without new harness — the arithmetic is the part that was wrong, and that's now covered.

GentleCold (approved):
LGTM

PR #370, opened 2026-06-26
fix(core): preserve per-slot NUMA on block rebuild to prevent 0-slot re-serve

xiaguan:
## Problem

A `SealedBlock` rebuilt from an RDMA fetch or SSD prefetch is constructed via
`SealedBlock::from_slots`, which left `slot_numas` empty even though `slots` was
fully populated. When such a block is later served to a peer,
`query_blocks_for_transfer` builds the wire slot list with:

```rust
block.slots().iter().zip(block.slot_numas())
```

`zip` stops at the shorter side, so an empty `slot_numas` truncates the served
slots to **zero**. The peer then rebuilds a 0-slot block and its load fails with:

```
stored block has 0 slots but instance ... expects N: namespace is shared by incompatible KV layouts
```

That message is misleading — the validator only sees `slots().len() == 0` and
blames layout. The real cause is a `slots` / `slot_numas` length desync, not a
namespace/layout collision.

## Root cause

`SealedBlock` kept `slots` and `slot_numas` as two parallel vecs that could
desync, and `from_slots` set `slot_numas = Vec::new()`. Both production rebuild
sites already had the real per-slot NUMA on hand but discarded it through
`from_slots`.

## Fix

Encode the invariant in the type: `from_slots` now takes
`Vec<(RawBlock, NumaNode)>` pairs, so a slot cannot be constructed without its
NUMA and `slot_numas.len() == slots.len()` holds by construction — a length
desync is no longer representable. Both rebuild sites thread the NUMA they
already had:

- RDMA fetch: the node its staging slab was allocated on (the same value already
  used for the local allocation), so a re-served block advertises where its
  bytes physically sit.
- SSD prefetch: `SlotMeta::numa_node`.

The SSD write path's empty-tolerant `debug_assert(slot_numas.is_empty() || ...)`
+ `.get(i).unwrap_or(UNKNOWN)` is tightened to a hard `assert_eq!` + direct
index, now that an empty `slot_numas` is unreachable.

Net effect: re-served fetched/SSD blocks keep all their slots **and** advertise
their true NUMA, so the next hop can stage NUMA-locally.

## Validation

- `cargo check` + `cargo clippy` clean
- 112 `pegaflow-core` lib unit tests pass
- 11 SSD roundtrip integration tests pass (exercise write → evict → rebuild)
- Adversarial review of NUMA correctness, the new `assert_eq!` reachability,
  `saturating_add` footprint behaviour, and `from_slots` callers — no issues

Merge-before P2P correctness E2E on a GPU host is still pending.

### DIFF
--- a/pegaflow-core/src/backing/rdma_fetch.rs
+++ b/pegaflow-core/src/backing/rdma_fetch.rs
@@ -315,6 +315,12 @@ async fn ensure_connected(
         .await
 }
 
+/// One fetched slot: its RDMA-staged segments plus the NUMA node they sit on.
+/// The NUMA travels with the slot so a re-served block advertises real topology.
+type StagedSlot = (Vec<SegmentAlloc>, NumaNode);
+/// A staged block awaiting SealedBlock rebuild: its hash and per-slot allocations.
+type StagedBlock = (Vec<u8>, Vec<StagedSlot>);
+
 /// Allocate local memory, build TransferDescs, execute RDMA READ, build SealedBlocks.
 async fn fetch_blocks_via_rdma(
     rdma: &RdmaTransport,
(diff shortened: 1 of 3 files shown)


GentleCold (approved):
LGTM

PR #380, opened 2026-07-01
fix(core): register only resident RDMA-prefetched blocks

GentleCold:
## Summary
- make ReadCache report which inserted blocks are actually resident after admission
- register RDMA-prefetched blocks to MetaServer only when they are resident in the local ReadCache
- keep SSD prefetch on the existing no-reregister path and add regression tests for LFU rejection/registration filtering

## Validation
- `cargo test -p pegaflow-core`
- `cargo clippy -p pegaflow-core --all-targets -- -D warnings`
- `cargo check -p pegaflow-core`

## Note
- `./scripts/check.sh` currently stops at Python ruff format because existing Python files would be reformatted; this PR only changes Rust core files.

### DIFF
--- a/pegaflow-core/src/cache.rs
+++ b/pegaflow-core/src/cache.rs
@@ -30,6 +30,7 @@ pub(crate) struct TinyLfuCache<K, V> {
     freq: Option<TinyLfu>,
 }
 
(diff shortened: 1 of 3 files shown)


xiaguan (approved):
LGTM

PR #382, opened 2026-07-03
fix(core): load blocks sealed by split-KV peers into contiguous device layouts

xiaguan:
## Problem

`build_copy_descs` pairs a `Contiguous` device block with only the host block's segment 0 for the full span. A block sealed by a peer with a **split-KV registration** (e.g. the vLLM connector's KV-first `(2, num_blocks, ...)` layout) stores K and V as **two separate host segments**, so a contiguous-layout instance loading it reads past the K allocation and restores garbage V data — silently.

This is the missing half of an existing asymmetry: the `Split` branch already falls back gracefully when the host block is contiguous (`v_ptr = k_ptr + k.bytes`), but the `Contiguous` branch had no handling for a split host block.

## Fix

When the host block carries two segments, emit one copy per segment targeting each half of the device span. Blocks whose segments don't exactly span the device block are rejected with an explicit error (same "incompatible KV layouts" family as the existing slot-count guard) instead of copying misaligned bytes.

Also carries a one-line `#[allow(too_many_arguments)]` on `rdma_v1::read_async_indices` — newer clippy fails the pre-commit hook on master; the signature mirrors the NIXL FFI 1:1.

## Validation

End-to-end P/D disaggregation on an 8-GPU H200 node (jz node 34):

- **P** = vLLM + PegaKVConnector (Qwen3-8B, `--block-size 16`, per-layer split-KV registration, 36 slots)
- **D** = openinfer decode instance (per-layer fused `[K|V]` pages, contiguous registration, 36 slots)
- 3-prompt smoke (short / cross-block / block-aligned), `temperature=0`: all outputs **byte-identical** to D's local-prefill baseline.
- RDMA fetch path exercised: 441 blocks / 992 MiB restored in 41.8 ms (23.7 GiB/s) for a 7k-token prompt.

Without this fix the same setup restores corrupted V segments.

### DIFF
--- a/pegaflow-core/src/gpu_worker.rs
+++ b/pegaflow-core/src/gpu_worker.rs
@@ -393,18 +393,46 @@ fn build_copy_descs(layers: &[LayerTransferData]) -> Result<(Vec<CopyDesc>, usiz
                     total_bytes += k.bytes + v.bytes;
                 }
                 BlockCopies::Contiguous(c) => {
-                    let ptr = block
-                        .block
-                        .raw()
-                        .segment_mapped_ptr(0)
-                        .unwrap()
-                        .add(host_offset);
-                    copies.push(CopyDesc {
-                        device: c.addr,
-                        host: ptr.host().as_ptr(),
-                        host_device: ptr.device().as_ptr() as u64,
-                        size: c.bytes,
-                    });
+                    let raw = block.block.raw();
+                    // Mirror of the Split branch's contiguous-host fallback:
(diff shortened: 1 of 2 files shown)


xiaguan:
Closing: the only consumer of this path was the qwen3 + legacy-vLLM P/D setup (split-KV host registration → contiguous-device load). The vLLM-prefill work has since settled on the vLLM 0.24+ GLM5.2 deployment (page-first MLA layout, which never takes this branch), and the legacy path is not a maintained target. Branch kept in case a split-KV producer ever returns.

Issue #383, opened 2026-07-06
P2P path hardening plan for store-based P/D disaggregation


xiaguan:
# P2P path hardening plan for store-based P/D disaggregation

## Context

The store-mediated P2P path (MetaServer discovery → RDMA READ fetch → local read cache) is intended to become the core of store-based prefill/decode disaggregation:

- **Prefill (write side)**: save → write pipeline seals blocks (`write_path.rs:288`) → fire-and-forget registration to MetaServer → `PegaEngine::flush_saves_and_registrations()` (`lib.rs:745`) → signal KV-ready.
- **Decode (read side)**: `query_prefetch` → local prefix scan → MetaServer query for the missing suffix → pick owner → `QueryBlocksForTransfer` (serving side pins blocks under a transfer lock) → RDMA READ into NUMA-local pinned slabs → rebuild `SealedBlock` into the read cache → re-register as a new owner → lease → H2D load.

#381 completed the skeleton: the flush barrier is the P/D handoff barrier, and `P2pTransferService` (`p2p_service.rs`) is the minimal serving surface for in-process embedders. The happy path works.

The gap: the path's semantics today are **best-effort cache** (any miss silently degrades to recompute). P/D disaggregation needs a **high-hit-rate handoff**. Every item below either closes a hit-rate hole or makes miss causes observable.

**Scope note**: transfer **v2 is out of scope** for this plan. The v1 READ engine (`RcBackend`) is the data plane throughout.

## Phase 1 — P/D readiness (correctness & observability)

Small, independent changes; one PR each.

### 1.1 Owner failover
`RdmaFetchStore::query_prefix` picks the single owner with the longest prefix (`rdma_fetch.rs:92-108`) and never retries another owner. If the chosen owner is dead or has evicted the blocks, the whole fetch fails to recompute — even though other owners are listed. Dead nodes stay query-visible for up to `node_stale_secs` (30s, `store.rs:433`), so a crashed prefill node causes a hard TTFT cliff for that window.

**Fix**: on fetch failure, retry the next owner by descending `prefix_len` before falling back to recompute.

### 1.2 Query-side metrics
`metrics.rs` covers registration/removal/heartbeat/rdma_fetch, but there are **no counters for MetaServer queries, hit rate, or miss attribution** (no-owner vs stale-owner vs transfer-error). The first production question for P/D will be "why is the remote hit rate low" — currently unanswerable.

**Fix**: add query count, hit/miss counters, and per-cause fetch-failure counters.

[...]

PR #384, opened 2026-07-06
refactor(transfer): narrow rc_backend hot-path lock and index MR lookups

xiaguan:
## What

Phase 2 of #383 — v1 RC backend (`rc_backend/`) hot-path cleanups. Four changes:

1. **Local MR lookup: linear scan → predecessor search.** `registered` moves from `HashMap` to a `LocalMemoryMap` (BTreeMap keyed by base_ptr); `find_mr` resolves any address with one predecessor lookup instead of scanning every registered region per descriptor. Registration now **rejects overlapping regions**, which makes predecessor lookup sound by construction — mirroring the validation the remote side already does on handshake snapshots. Same-base duplicate registration is also rejected: silently replacing would dereg an MR whose rkey peers already hold from a handshake, turning their next READ into a remote access error. The overlap error carries both regions' base/len for diagnosis.
2. **`batch_transfer_async` no longer holds the global state mutex across op building.** Previously the lock covered the whole per-descriptor MR/rkey lookup loop, serializing concurrent fetches to different peers. Now a short lock snapshots Arc'd state — the local memory map (RCU-style clone-and-swap on the rare register/unregister), the per-connection `RemoteMemorySnapshot`, and the session vec — and all per-descriptor lookups run outside the lock. Lifetime safety is unchanged: ops already held `Arc<MemoryRegion>`/`Arc<RcSession>` past the lock on master; the snapshots only extend how long a concurrently-deregistered MR stays alive, never shorten it.
3. **Removed the per-session recv CQ.** The path is one-sided RDMA only — no recvs are ever posted, no IMM, no SRQ — so no recv completion can exist. The send CQ is passed as the recv CQ at QP creation.
4. **Hot-path `.expect("session vec must exist…")` is now an error return**, degrading the fetch to recompute instead of panicking the caller on inconsistent state.

Also includes a standalone commit adding `#[allow(clippy::too_many_arguments)]` to `read_async_indices` (pre-existing 8-arg PyO3 binding flagged by clippy 1.95, blocks any commit via pre-commit).

## Behavior changes

- `register_memory` now returns an error for overlapping or duplicate-base registrations (previously: silent `HashMap::insert` replacement). Current callers register disjoint regions once (pinned-pool shards; deduped vLLM KV bases), so this should only fire on genuinely broken layouts — loudly, at startup, with both regions in the message.
- A missing session/remote-memory entry for an established connection is now a returned `Backend` error instead of a panic.

## Not claimed

No performance numbers in this PR: this machine has no RDMA fabric, so the lock-scope and lookup changes are verified by unit tests and review only. Framed as a correctness/scalability cleanup, not a measured win.

## Testing

- [x] `cargo test -p pegaflow-transfer` (17 passed) — includes new `LocalMemoryMap` cases: overlap with predecessor/successor rejected, same-base rejected, adjacent regions allowed, zero-length rejected, covered/straddling lookup
- [x] `cargo clippy -p pegaflow-transfer -p pegaflow-core --all-targets` clean, `cargo fmt --check` clean, pre-commit hooks pass
- [x] Toxic-review pass: concurrency equivalence vs master traced (unregister/invalidate races, Arc lifetimes, CQ destruction order); review findings applied (overlap error detail, same-base rejection, `len > 0` invariant inside `insert`, adjacent-boundary test, redundant bound check removed)
- [x] Real-link A/B on an 8× 400G IB node: `cpu_bench` master vs PR across large-set / single-block / descriptor-heavy / 4-NIC shapes — no regression, differences ≤3% within noise (table in PR comment); crate tests pass on the real-NIC node
- [ ] **Before merge**: vLLM e2e for the connector-visible `register_memory` rejection path (test node GPUs were fully occupied)

### DIFF
--- a/pegaflow-transfer/src/rc_backend/mod.rs
+++ b/pegaflow-transfer/src/rc_backend/mod.rs
@@ -13,7 +13,7 @@ use sideway::ibverbs::AccessFlags;
 
 use self::runtime::RcRuntime;
 use self::session::{RcSession, RdmaOp};
-use self::state::{AddrConnection, RcBackendState, RegisteredMemoryEntry};
+use self::state::{AddrConnection, RcBackendState, RegisteredMemoryEntry, RemoteMemorySnapshot};
 use std::ptr::NonNull;
 
 use mea::oneshot;
@@ -163,14 +163,11 @@ impl RcBackend {
         }
 
         let mut state = self.state.lock();
-        state.registered.insert(
-            raw,
-            RegisteredMemoryEntry {
-                base_ptr: raw,
-                len,
-                mrs,
-            },
-        );
+        Arc::make_mut(&mut state.registered).insert(RegisteredMemoryEntry {
+            base_ptr: raw,
+            len,
+            mrs,
+        })?;
         debug!(
             "memory registered: ptr={:#x}, len={}, nics={}",
             raw,
@@ -183,7 +180,7 @@ impl RcBackend {
     pub(crate) fn unregister_memory(&self, ptr: NonNull<u8>) -> Result<()> {
         let raw = ptr.as_ptr() as u64;
         let mut state = self.state.lock();
-        let removed = state.registered.remove(&raw);
+        let removed = Arc::make_mut(&mut state.registered).remove(raw);
         if removed.is_none() {
             return Err(TransferError::MemoryNotRegistered { ptr: raw });
         }
@@ -196,17 +193,16 @@ impl RcBackend {
         let state = self.state.lock();
         (0..self.nic_count())
             .map(|nic_idx| {
-                let mut regions: Vec<RegisteredMemoryRegion> = state
+                // LocalMemoryMap iterates in base_ptr order already.
+                state
                     .registered
-                    .values()
+                    .iter()
                     .map(|entry| RegisteredMemoryRegion {
                         base_ptr: entry.base_ptr,
                         len: entry.len as u64,
                         rkey: entry.mrs[nic_idx].rkey(),
                     })
-                    .collect();
-                regions.sort_unstable_by_key(|entry| entry.base_ptr);
-                regions
+                    .collect()
             })
             .collect()
     }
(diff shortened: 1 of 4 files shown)


xiaguan:
## Real-link A/B validation

Ran `pegaflow-cpu-bench` (pure CPU memory + ibverbs, no CUDA) on an idle-NIC H200 node with 8× 400G IB, master (`d46fd16`) vs this branch, same node, interleaved runs, `--mode read` (the store fetch path). The bench exercises the full changed surface in-process: `register_memory` (new overlap-validated `LocalMemoryMap`), prepare/complete handshake (`RemoteMemorySnapshot`), and `batch_transfer_async` (short-lock + outside-lock lookups, shared-CQ sessions).

| Shape | master p50 | PR p50 | note |
|---|---|---|---|
| 150 × 4MB, 1 NIC | 12.07 ms (48.54 GiB/s) | 12.07 ms (48.55 GiB/s) | line rate, identical |
| 1 × 4MB, 1 NIC, 300 tasks | 0.12 ms | 0.12 ms | identical |
| 512 × 64KB, 1 NIC (descriptor-heavy) | 0.71 ms (44.0 GiB/s) | 0.69 ms (45.1 GiB/s) | PR ~3% better |
| 150 × 4MB, 4 NICs aggregate | 5.92 / 5.73 ms (98.9 GiB/s) | 5.79 / 5.69 ms (102.1 GiB/s) | 2 runs each; PR ≤ master in every pairing, ranges overlap |

Conclusion: **no regression on real 400G link across all shapes; differences are within noise (≤3%)**. The lock-contention benefit (concurrent fetches to different peers) is not measurable with this single-peer bench and remains unclaimed. `cargo test -p pegaflow-transfer --release` also passes on the real-NIC node (17/17).

Still pending before merge: vLLM e2e for the connector-visible `register_memory` rejection path (GPUs on the test node were fully occupied; the bench node's registration + handshake path did run the new validation code successfully in every run above).

GentleCold (approved):
LGTM

PR #404, opened 2026-07-17
feat(cache): prioritize reclaiming redundant KV replicas

GentleCold:
## Summary

Related RFC: #398

Reduce redundant resident KV copies in P/D deployments with a two-class replacement policy:

- Locally saved blocks enter `retained`.
- RDMA/P2P-fetched blocks enter `reclaimable`.
- P-side `finished_sending` marks the saved blocks for that request as `reclaimable`.
- MetaServer Insert returns aligned post-insert owner counts; a newly saved local block is asynchronously marked `reclaimable` when `owners >= 3`.
- Reclaim evicts `reclaimable` blocks before `retained` blocks, with LRU order inside each class.
- Already-existing inserts, local prefix hits, and cross-node serving hits refresh recency without changing class.

The policy changes replacement priority only. It does not change NIXL, add a data-path RPC, delay registration, or introduce backpressure.

## PD E2E Validation

Baseline and feature used the same generated multi-turn corpus and seed in a single-host 2P2D deployment:

- Four independent TP2 vLLM instances: 2 prefill and 2 decode.
- P side: Pega read/write + NIXL pull through vLLM `MultiConnector`.
- D side: NIXL pull + Pega save-only.
- vLLM PD proxy with round-robin P/D selection.
- vLLM internal prefix cache disabled.
- Pega huge pages enabled; pool size `150 GB` per instance.
- Model: `DeepSeek-V2-Lite-Chat`.
- 512 conversations, 16 messages / 8 model requests per conversation, 4096 requests total.
- 8192 input tokens per generated user message and 128 output tokens per request.
- 32 clients, 32 active conversations, closed-loop with no request-rate limit, seed `0`.
- MetaServer metrics sampled every 5 seconds with no smoothing.
- The generated input corpus was identical in both runs: SHA256 `e3443b744287912e79e2de3a72cfbfade82972a9ec4eed77b94f5b94877ef0d1`.

Both runs completed all `4096/4096` requests. The pool was large enough to represent the intended deployment shape while the sustained workload still created eviction pressure.

### Results

| Metric | Baseline | Feature | Change |
| --- | ---: | ---: | ---: |
| Final average redundancy | 2.9278 | 2.1661 | -26.02% |
| Whole-run non-empty mean redundancy | 2.7939 | 2.1605 | -22.67% |
| Tail 20% mean redundancy | 2.8886 | 2.1106 | -26.93% |
| Final blocks with owners >= 4 | 106,188 | 3,635 | -96.58% |
| Final live owner records | 687,782 | 657,385 | -4.42% |
| Evicted blocks | 6,773,760 | 8,646,656 | +27.65% |
| RDMA fetch bytes | 799.00 GB | 787.09 GB | -1.49% |
| Cache-miss blocks | 2,127,355 | 2,127,355 | unchanged |
| Request throughput | 1.226 req/s | 1.223 req/s | -0.24% |
| Mean TTFT | 25,512.45 ms | 25,521.10 ms | +8.65 ms |

PD 2P2D redundancy comparison: baseline vs replica-aware reclamation

The feature substantially suppresses high-replication blocks once reclaim pressure begins. It performs more evictions, but the identical cache-miss count, slightly lower RDMA volume, and effectively unchanged throughput/TTFT indicate that this run did not turn the additional eviction into extra recomputation or transfer cost.

## Scope and Tradeoffs

- Owner hints are best-effort replacement signals, not correctness state.
- The `owners >= 3` rule applies to every newly inserted local resident, including D-side save-only copies; the policy does not pin all D-side copies as `retained`.
- More aggressive reclamation increases eviction churn. Metrics for class distribution and evictions by class are included so this can be monitored after rollout.
- Pega Server and MetaServer are upgraded together; a successful Insert response with missing or misaligned `owner_counts` is treated as a protocol error.

## Checks

- Rust formatting and clippy passed with warnings denied.
- `pegaflow-core`: 139 passed, 1 ignored.
- `pegaflow-metaserver`: 31 passed.
- Python connector tests: 38 passed.
- Release server, MetaServer, and PyO3 builds passed in the remote validation environment.
- Baseline and feature both completed the 2P2D workload.

### DIFF
--- a/pegaflow-core/src/backing/transfer_lock_guard.rs
+++ b/pegaflow-core/src/backing/transfer_lock_guard.rs
@@ -85,7 +85,8 @@ mod tests {
         QueryBlocksForTransferResponse, QueryRequest, QueryResponse, RdmaHandshakeRequest,
         RdmaHandshakeResponse, RegisterContextRequest, RegisterContextResponse, ReleaseRequest,
         ReleaseResponse, ReleaseTransferLockResponse, SaveRequest, SaveResponse, SessionEvent,
-        SessionRequest, ShutdownRequest, ShutdownResponse, UnregisterRequest, UnregisterResponse,
+        SessionRequest, SetColdBlocksRequest, SetColdBlocksResponse, ShutdownRequest,
+        ShutdownResponse, UnregisterRequest, UnregisterResponse,
     };
     use tokio_stream::wrappers::TcpListenerStream;
     use tonic::transport::Endpoint;
@@ -159,6 +160,12 @@ mod tests {
             &self,
             _request: Request<UnregisterRequest>,
         ) -> Result<Response<UnregisterResponse>, Status> {
+            Err(Status::unimplemented("stub"))
+        }
+        async fn set_cold_blocks(
+            &self,
+            _request: Request<SetColdBlocksRequest>,
+        ) -> Result<Response<SetColdBlocksResponse>, Status> {
             Err(Status::unimplemented("stub"))
         }
         async fn shutdown(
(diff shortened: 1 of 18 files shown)


GentleCold:
Superseded by #412, which narrows the policy to source-based retained/reclaimable admission without owner hints, P-side demotion, hit promotion, or connector changes.

PR #409, opened 2026-07-20
fix(core): drain GPU workers before unregister

xiaguan:
## Summary

- reject new GPU transfers and drain both worker streams before removing an instance
- synchronize CUDA streams at the drain barrier, including transfer error paths
- make server cleanup wait for the engine drain before releasing registered CUDA tensors
- serialize cleanup against registration and session creation, and keep cleanup running after caller cancellation
- preserve the synchronous unregister APIs while adding drained variants for allocation owners

## Testing

- `prek run`
- core unit tests: 127 passed, 1 ignored
- server unit tests: 23 passed
- HTTP cleanup integration test: 1 passed


xiaguan:
Folded into #407 so native CUDA IPC registration and the required drain-before-unmap lifecycle are reviewed and merged together.

PR #410, opened 2026-07-20
docs: add NIXL P/D deployment example

xiaguan:
## Summary

- add a GLM-5.2 FP8 P/D deployment example using MTP and MultiConnector on both sides
- configure NIXL as the P-to-D transport and PegaFlow as read/write on prefill and save-only on decode
- document shared hash/layout invariants and D-to-P reuse validation
- link the guide from the main README

## Validation

- `prek run --files README.md docs/deployment.md`
- parsed both connector examples as JSON
- validated child connector configurations with vLLM `KVTransferConfig`
- checked embedded shell snippets with `bash -n`

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -2,69 +2,19 @@
 
 ## P/D with NIXL
 
-This example runs GLM-5.2 FP8 with prefill/decode disaggregation, MTP, and
-vLLM's `MultiConnector` on both sides.
+This GLM-5.2 FP8 example enables MTP and `MultiConnector` on both sides. NIXL
+handles P-to-D transfer; PegaFlow uses `read_write` on P and `save_only` on D
+to reuse KV produced by earlier requests without competing with NIXL loads.
 
-```text
-request -> prefill -- NIXL --> decode
-              |                   |
-              +---- PegaFlow -----+
-```
-
-NIXL transfers the current request's KV cache from prefill to decode.
-PegaFlow stores reusable KV cache across requests and nodes:
-
-- The prefill instance uses PegaFlow in `read_write` mode. It can reuse a
-  cached prefix, including output KV previously saved by a decode instance.
-- The decode instance puts NIXL first and uses PegaFlow in `save_only` mode.
-  NIXL owns the current P-to-D load, while PegaFlow persists the resulting KV
-  cache without competing for that load.
-
-### Prerequisites
-
-- Run a PegaFlow server beside each vLLM instance and connect those servers to
-  the same PegaFlow metaserver.
-- Configure NIXL networking and a P/D-aware router for the deployment
-  environment.
-- Use the same model revision, tokenizer, block size, KV cache dtype,
-  `PYTHONHASHSEED`, and compatible KV cache layout on both sides.
-- Set `PYTHONHASHSEED` explicitly. Different values produce different vLLM
-  block hashes and prevent cross-instance cache reuse.
-
-The following commands assume the local PegaFlow server listens on
-`http://127.0.0.1:50055`. Parallelism and memory limits are examples; adjust
-them together with the router configuration for the target hardware.
+Run one PegaFlow server beside each vLLM instance, backed by the same metaserver,
+and use a P/D-aware NIXL router. P and D must use compatible model, tokenizer,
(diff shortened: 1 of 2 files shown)


xiaguan:
Closing because deployment documentation is not part of the current external PegaFlow delivery scope.

docs/deployment.md
```diff
@@ -0,0 +1,104 @@
...
+
+Run one PegaFlow server beside each vLLM instance, backed by the same metaserver,
+and use a P/D-aware NIXL router. P and D must use compatible model, tokenizer,
+block size, KV dtype, KV layout, and `PYTHONHASHSEED`. The example uses seed 42
+and a local PegaFlow server at `http://127.0.0.1:50055`.
+
+### Prefill
+
+```bash
+PYTHONHASHSEED=42 \
```
GentleCold, line 24: **[P1] Configure the NIXL side channel explicitly**

Both commands currently inherit `VLLM_NIXL_SIDE_CHANNEL_HOST=localhost` and `VLLM_NIXL_SIDE_CHANNEL_PORT=5600`. If P and D run on the same host, the two NIXL schedulers contend for port 5600; if they run on different hosts, each side advertises `localhost`, so its peer cannot reach it. In either topology the documented P-to-D path does not work as written. Please set a routable side-channel host on each node and use distinct ports when both instances share a host; the decode command needs the corresponding settings as well.

docs/deployment.md
```diff
@@ -0,0 +1,104 @@
...
+        {
+          "kv_connector": "NixlConnector",
+          "kv_role": "kv_producer"
+        },
+        {
+          "kv_connector": "PegaKVConnector",
+          "kv_role": "kv_both",
+          "kv_connector_module_path": "pegaflow.connector",
+          "kv_connector_extra_config": {
+            "pegaflow.host": "http://127.0.0.1",
```
GentleCold: **[P1] Use a routable PegaFlow endpoint for cross-node reuse**

The guide requires one PegaFlow server beside each vLLM instance, with both servers connected through the metaserver, but this connector is pinned to loopback (and the decode config does the same). PegaFlow uses `--addr` as both its bind address and the address advertised to peers: binding a routable node IP leaves no listener on `127.0.0.1`, while binding loopback makes the server unreachable to the other PegaFlow node. As a result, the D-to-P `query_blocks_for_transfer`/RDMA path described below cannot work. Please use a routable PegaFlow address for each node here and document or link the matching `--addr`, `--nics`, and `--metaserver-addr` server setup.

docs/deployment.md
```diff
@@ -0,0 +1,113 @@
...
+
+Run one PegaFlow server beside each vLLM instance, backed by the same metaserver,
+and use a P/D-aware NIXL router. Configure each server's routable `--addr`,
+`--nics`, and `--metaserver-addr` as described in the [P2P guide](./p2p.md).
+P and D must use compatible model, tokenizer, block size, KV dtype, KV layout,
+and `PYTHONHASHSEED`.
+
+The commands use the documentation addresses `192.0.2.10` for P and
+`192.0.2.11` for D; replace them with addresses assigned to the target nodes.
+The distinct NIXL ports also avoid a collision if P and D are colocated.
```
GentleCold: **[P2] Keep the colocated setup internally consistent**

Distinct NIXL ports avoid only the NIXL side-channel collision. In a colocated deployment these documentation addresses will normally be replaced by the same host address, but both PegaFlow connectors still target port `50055`. Since the guide requires one PegaFlow server beside each vLLM instance, the second server cannot bind that same host/port (or both connectors would unintentionally use one server), so the documented topology is still not colocatable as stated. Please either scope this example to separate P/D nodes, or assign the decode-side PegaFlow server and connector a different port as well.

xiaguan: Sharing one PegaFlow server between colocated P and D is intentional and supported: each vLLM instance registers with its own instance id (vLLM engine_id/instance_id, or a generated UUID fallback), so both connectors can point at the same address without ambiguity — and with both sides local, reuse works without the metaserver/RDMA path.

Updated in 8e59600: the example is scoped to separate P/D nodes, and the colocation note now says to keep the distinct NIXL side-channel ports and point both connectors at a single shared PegaFlow server.

GentleCold (approved):
LGTM

PR #411, opened 2026-07-20
feat: complete remote prefetch with partial tail blocks

xiaguan:
## Summary

- add `pegaflow.wait_for_full_prefix` so a query stays in PegaFlow's prefetch state until the remaining prefix is complete
- load prefill-saved partial prompt blocks on decode while recomputing the logits token
- make CUDA IPC registration select and synchronize the correct device
- document the P/D hash seed, hash algorithm, and connector settings

## Validation

- `prek run`
- `cd python && uv run --extra test pytest` (199 passed, 11 deselected)
- source-only Python gate (199 passed, 11 deselected)
- GLM-5.2 FP8 on H200 x8: a new 12,020-token prompt had zero cold hits; after prefill saved it, decode loaded 188 blocks through PegaFlow, reported `tail_hit=True`, reused 12,019/12,020 prompt tokens, and produced the same greedy next token as the cold baseline
- four 4-turn conversations completed 16/16 requests; PegaFlow recorded 2,836 block hits, zero new misses, and 16/16 partial-tail hits as context grew from about 8K to 14K tokens
- eight random 7K-10K prompts completed 8/8 requests with 1,056 block hits, zero new misses, and 8/8 partial-tail hits

## Performance Work

- scheduler-side cache lookup was 0.43-0.71 ms in the multi-turn run
- no TTFT improvement is claimed: the correctness harness serializes a complete prefill request before decode, so its end-to-end TTFT is not representative of a production P/D router

## Known Behavior

- cancelling a connector probe does not cancel an already-running server-side remote wait; that task is bounded by the existing 30-second timeout and stale-entry GC

### DIFF
--- a/pegaflow-core/src/storage/prefetch.rs
+++ b/pegaflow-core/src/storage/prefetch.rs
@@ -53,7 +53,17 @@ impl RdmaFetch {
             .fetch_blocks(&node, req_id, namespace, &remaining_hashes[..found])
             .await;
         if require_full_prefix && blocks.len() != found {
-            return None;
+            // The advertised owner served fewer blocks than the MetaServer
+            // promised (stale advertisement or failed fetch). Keep the partial
+            // result so poll_existing blacklists RDMA for this request instead
+            // of the wait loop retrying the same fetch until timeout.
+            warn!(
+                "RDMA fetch returned fewer blocks than advertised: req_id={} node={} returned={} advertised={}",
+                req_id,
+                node,
+                blocks.len(),
+                found
+            );
         }
         Some((found, blocks))
     }
@@ -141,8 +151,9 @@ struct PrefetchState {
     active: HashMap<String, PrefetchEntry>,
     /// Reserved SSD prefetch budget for active background tasks.
     reserved_ssd_prefetch_blocks: usize,
-    /// req_ids where RDMA remote fetch returned zero blocks (remote evicted).
-    /// Prevents re-triggering RDMA on every subsequent poll for the same request.
+    /// req_ids where the advertised RDMA owner served fewer blocks than the
+    /// MetaServer promised (stale advertisement or failed fetch). Prevents
+    /// re-triggering RDMA on every subsequent poll for the same request.
     failed_remote: HashMap<String, Instant>,
 }
 
@@ -464,6 +475,7 @@ fn reserve_ssd_prefetch_slots(
     state: Arc<Mutex<PrefetchState>>,
     max_prefetch_blocks: usize,
     requested: usize,
+    require_full: bool,
 ) -> Option<(usize, SsdPrefetchReservation)> {
     if requested == 0 {
         return None;
@@ -472,7 +484,7 @@ fn reserve_ssd_prefetch_slots(
     let mut guard = state.lock();
     let available = max_prefetch_blocks.saturating_sub(guard.reserved_ssd_prefetch_blocks);
 
-    if available == 0 {
+    if available == 0 || (require_full && available < requested) {
         core_metrics()
             .ssd_prefetch_backpressure_blocks
             .add(requested as u64, &[]);
@@ -582,11 +594,15 @@ async fn run_prefetch_task(deps: PrefetchTaskDeps, input: PrefetchTaskInput) -> 
                 Arc::clone(&deps.prefetch_state),
                 deps.max_prefetch_blocks,
                 found,
+                wait_for_full_prefix,
             )
         {
             let keys = remaining_keys[..reserved].to_vec();
             let (found, blocks) = ssd.prefetch_prefix(keys).await;
-            if found > 0 {
+            // wait_for_full_prefix is all-or-nothing: a partial SSD result
+            // (backpressured reservation or short read) must not let the
+            // caller proceed with a partial prefix.
+            if found > 0 && (!wait_for_full_prefix || found == remaining_keys.len()) {
                 record_tier_attribution(
                     total,
                     hit,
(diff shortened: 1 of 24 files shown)


pegaflow-core/src/storage/prefetch.rs
```diff
@@ -39,12 +42,19 @@ impl RdmaFetch {
...
     ) -> Option<(usize, PrefetchResult)> {
         let (node, found) = self.0.query_prefix(namespace, remaining_hashes).await?;
+        if require_full_prefix && found != remaining_hashes.len() {
+            return None;
+        }
         let blocks = self
             .0
             .fetch_blocks(&node, req_id, namespace, &remaining_hashes[..found])
             .await;
+        if require_full_prefix && blocks.len() != found {
```
GentleCold, line 55: [P1] Do not collapse a stale owner into the normal wait retry path. When MetaServer reports the full prefix but fetch_blocks() returns fewer blocks, this returns None; the loop at lines 611-615 then retries the full RDMA fetch every 10ms for up to 30s. Because the partial result is discarded, poll_existing() never sees source=Rdma with cache_inserts.len() < found, so failed_remote is not populated. A stale owner or mid-fetch allocation failure can therefore create thousands of repeated MetaServer/RDMA attempts. Please distinguish "producer not registered yet" from "advertised owner cannot serve the full prefix", and back off or blacklist the latter.

python/pegaflow/connector/scheduler.py
```diff
@@ -506,6 +546,45 @@ def _consume_tail_save(self, req_id: str, written: int) -> SaveIntent | None:
         )
         return SaveIntent(block_ids=(allocated[tail_idx],), block_hashes=(tail_key,))
 
+    def _derive_tail_block(self, request: "Request") -> tuple[bytes, int] | None:
+        if self._tail_hash_fn is None:
+            return None
+        # The tail key carries no extra_keys. Reusing it for salted, LoRA, or
+        # multimodal requests would alias distinct vLLM cache identities.
+        if request.lora_request is not None or request.cache_salt or request.mm_features:
```
GentleCold, line 554: [P1] Guard prompt-embedding requests before deriving the tail hash. Request.prompt_token_ids can be None when prompt_embeds is used, so line 572 raises TypeError as soon as pd_tail_load is enabled. For mixed token/embedding requests, the code does not crash but the tail key omits vLLMs prompt-embedding extra key, so different embeddings can alias to the same KV tail. Please disable tail save/load for any request with prompt embeddings, or derive the exact same extra keys as vLLM.

pegaflow-core/src/storage/prefetch.rs
```diff
@@ -552,39 +575,69 @@ async fn run_prefetch_task(deps: PrefetchTaskDeps, input: PrefetchTaskInput) ->
...
+                    Some(PrefetchSource::Ssd),
+                    found,
+                    &remaining_keys[..found],
+                    blocks,
+                );
+            }
         }
     }
 
+    if wait_for_full_prefix && let Some(rdma) = deps.rdma_fetch {
```
GentleCold, line 625: [P1] wait_for_full_prefix does not observe writes to the local/shared PegaEngine. The wait loop only runs when deps.rdma_fetch exists and only re-queries MetaServer; it never rescans ReadCache (and the SSD path is checked only once). With the shared PegaEngine topology documented in docs/pd.md, a Prefill save that completes after this query is invisible, and without RDMA/SSD start_prefetch_task() returns false immediately. Either make the wait condition observe local cache changes, or explicitly scope and document this option as RDMA-only.

pegaflow-core/src/storage/prefetch.rs
```diff
@@ -552,39 +575,69 @@ async fn run_prefetch_task(deps: PrefetchTaskDeps, input: PrefetchTaskInput) ->
...
-            return build_ready_result(prefix_blocks, total, None, 0, &[], Vec::new());
-        };
-        let keys = remaining_keys[..reserved].to_vec();
-        let (found, blocks) = ssd.prefetch_prefix(keys).await;
-        if found > 0 {
-            record_tier_attribution(
-                total,
-                hit,
+        if (!wait_for_full_prefix || found == remaining_keys.len())
+            && let Some((reserved, _reservation)) = reserve_ssd_prefetch_slots(
```
GentleCold, line 593: [P2] Strict full-prefix mode can still return a partial SSD result under backpressure. found == remaining_keys.len() confirms that SSD has the whole prefix, but reserve_ssd_prefetch_slots() may return reserved < found; the code then fetches remaining_keys[..reserved] and returns Ready on any positive result. That violates the wait_for_full_prefix contract and lets Decode proceed with a partial prefix. In this mode, require reserved == found, or keep the request Loading until capacity is available.

GentleCold (commented):
Review findings: four issues below need to be addressed before merge.

python/pegaflow/connector/scheduler.py
xiaguan: Prompt-embedding requests are outside the supported scope of the P/D tail extension. This feature targets token-ID prompts used by our OpenAI-compatible P/D serving path. We intentionally do not derive prompt-embedding tail identities or silently fall back for this unsupported input.

xiaguan:
@GentleCold All four findings are addressed:

- [P1] Stale owner: short fetches now keep the partial result and blacklist the request from further RDMA attempts, instead of collapsing to `None` and retrying the full fetch every 10ms for 30s. Contract tests added.
- [P1] `wait_for_full_prefix` scope: now explicitly documented as remote-producer-only (MetaServer + RDMA) in the pyi stub, proto, README, and docs/pd.md.
- [P2] SSD strict mode: reservation is all-or-nothing under backpressure, and short SSD reads no longer return a partial prefix. Contract test added.
- [P1] prompt_embeds: replied in the original thread — outside the supported scope of the P/D tail feature.

Fixed and pushed. Please review again.

GentleCold (approved):
LGTM

Issue #389, opened 2026-07-09
Proposal: NIXL pull-aware eviction demotion for P-side Pega blocks


GentleCold:
## Background

Current production PD-separated deployments can run vLLM `MultiConnector` with multiple KV connectors, for example:

- P side: `PegaKVConnector` + NIXL pull connector
- D side: NIXL pull connector + `PegaKVConnector(save_only)`

NIXL pull handles D-side RDMA READ from P-side KV. D-side Pega runs in save-only mode so decode-side host memory can be used as a cache tier.

This creates a redundancy problem for transferred prompt blocks:

```text
P-side Pega persistent copy
D-side Pega save-only persistent copy
D-side runtime/GPU KV copy
```

The effective redundancy can become greater than 2 for the same block payload. Disabling D-side Pega save-only is not acceptable because it wastes D-side memory and loses the intended decode-side cache tier.

## Proposal

Add a **NIXL pull-aware eviction demotion hint** for P-side Pega blocks.

Instead of hard-moving ownership or disabling P-side save, keep the current data flow and make P-side blocks that have already been pulled by D easier to evict under memory pressure.

```text
P side:
  Pega saves blocks normally
  NIXL exposes blocks for D-side READ and waits for D completion notification
  after the request's async Pega save and NIXL pull notification complete:
    demote the corresponding Pega blocks in the replacement policy

D side:
  NIXL pulls KV from P normally
  Pega save_only stores cache blocks normally
```

This preserves short-term P-side reuse when memory is available, but under pressure the duplicate P-side copy is reclaimed before normal blocks.

## Why this fits vLLM MultiConnector + NIXL pull

vLLM `MultiConnector` already aggregates child connector completion:

- `get_finished()` merges `finished_sending` / `finished_recving` from all child connectors.
- For multiple async send/save connectors on the same request, `extra_async_saves` delays aggregated `finished_sending` until all async children have completed.
- `update_connector_output()` then broadcasts the same aggregated output to every child connector scheduler.

For NIXL pull specifically:

- D side completes the RDMA READ and sends a notification back to the P-side NIXL worker.
- P-side NIXL reports that notification as `finished_sending` for the original P-side request after all expected D consumers have read the blocks.
- In `MultiConnector(Pega + NIXL pull)`, aggregated `finished_sending` means P-side Pega save has completed and NIXL has observed D-side pull completion for that request.
- Pega can use this aggregated `connector_output.finished_sending` as the demotion trigger.

[...]

PR #413, opened 2026-07-22
feat(cache): reclaim extra replicas with MetaServer hints

GentleCold:
## Summary

- make MetaServer decide which newly registered replicas are reclaimable: a hash is returned only when this node becomes a new third-or-later owner
- return self-describing `reclaimable_hashes` from ordinary insert RPCs instead of aligned `owner_counts`
- apply insert hints in the shared registration loop through `Weak<ReadCache>`, moving matching retained residents to the reclaimable LRU
- remove `InsertWithHint`, per-batch callbacks, owner-count alignment checks, and compatibility handling for the old response contract
- preserve the source policy from #412: RDMA/P2P residents are reclaimable, local/SSD residents start retained, and `AlreadyExists` does not otherwise change class

This builds on #412. Related design/background: #398 and #389.

## Policy

```text
Local/SSD insert -> retained
RDMA/P2P fetch   -> reclaimable

MetaServer insert:
  existing owner                 -> no hint
  new owner, post-insert count <3 -> no hint
  new owner, post-insert count >=3 -> return hash in reclaimable_hashes

Pega registration loop:
  matching retained resident -> reclaimable
  missing/already reclaimable -> no-op

Eviction order: reclaimable LRU -> retained LRU
```

`reclaimable_hashes` is an insert-time signal, not a continuously synchronized replica count. Pega server and MetaServer are deployed at the same version, so the protobuf field is replaced directly without a legacy compatibility path. Registration remains best effort; a lost hint can only affect replacement pressure, not KV correctness or owner registration/removal.

## 2P6D Benchmark

The latest run uses the same workload and result files for all three policies.

| Setting | Value |
| --- | --- |
| Model | DeepSeek-V2-Lite-Chat |
| Topology | 2 prefill + 6 decode, all TP1; 8 independent Pega servers |
| Connectors | P: Pega read/write + NIXL producer; D: NIXL consumer + Pega save-only |
| Proxy | official vLLM non-stream multi-turn P/D proxy |
| Pega pool | 75 GB per server, hugepages enabled |
| Workload | 256 conversations, 16 turns, 2048 completed requests |
| Clients | 32 closed-loop clients, 32 active conversations |
| Tokens | 8192 input per turn / 1024 output; max model length 81920 |
| Sampling | round-robin conversations, seed 0, request rate 0 |
| Metrics | 5-second raw sampling |

vLLM prefix caching and the proxy conversation-metadata cache were disabled, matching the previous baseline/source-tiered runs. This keeps conversation-history reuse on the Pega P2P path.

### Redundancy and performance

| Policy | Mean redundancy | Tail 20% | Final | Runtime | Req/s | Mean TTFT |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Baseline | 2.961 | 2.890 | 3.143 | 3923 s | 0.522 | 58453 ms |
| Two-tier source policy | 3.029 | 3.089 | 3.574 | 3875 s | 0.529 | 58208 ms |
| **Two-tier + MetaServer hint** | **1.736** | **1.359** | **1.409** | **3901 s** | **0.525** | **58451 ms** |

Relative to baseline, MetaServer hints reduced mean redundancy by **41.4%**, the tail-20% mean by **53.0%**, and final redundancy by **55.2%**. Throughput and mean TTFT remained effectively unchanged.

### Final owner distribution

| Policy | Owner=1 | Owner=2 | Owner=3 | Owner>=4 |
| Baseline | 59,146 | 31,837 | 23,677 | 90,188 |
| Two-tier source policy | 31,963 | 25,633 | 26,705 | 94,386 |
| **Two-tier + MetaServer hint** | **293,918** | **138,022** | **9,051** | **8,008** |

The high-owner population collapses after reclaim starts: final owner>=4 blocks fall by 91.1% relative to baseline. The larger owner=1/2 populations are the expected result of preferentially reclaiming third-or-later replicas.

### Reclamation cost

| Policy | Evictions | RDMA fetches | RDMA bytes | Cache miss blocks |
| Baseline | 5,386,752 | 902 | 615,231,645,696 | 1,164,029 |
| Two-tier source policy | 6,294,016 | 1,452 | 1,453,582,485,504 | 1,164,605 |
| **Two-tier + MetaServer hint** | **6,893,568** | **1,463** | **925,819,269,120** | **2,349,885** |

The reduction is not free: versus baseline, evictions increased 28.0%, RDMA bytes increased 50.5%, and cache-miss blocks increased 101.9%. The result demonstrates that the policy can trade additional refetch pressure for substantially lower redundancy; production adoption should therefore consider network and latency budgets alongside the redundancy metric.

2P6D baseline, source-tiered, and MetaServer-hint redundancy comparison

## Previous Topology Results

The earlier 2P2D and 4-mixed runs used 150 GB/server, TP2, 512 conversations, 16 turns, 8192 input / 128 output, 32 clients, seed 0, and 5-second sampling.

| Topology / policy | Mean redundancy | Tail 20% | Final | Runtime | Req/s |
| --- | ---: | ---: | ---: | ---: | ---: |
| 2P2D baseline | 2.798 | 2.889 | 2.928 | 3342 s | 1.226 |
| 2P2D two-tier source policy | 2.215 | 2.185 | 2.136 | 3386 s | 1.210 |
| **2P2D two-tier + owner hint** | **1.961** | **1.814** | **1.884** | **3324 s** | **1.232** |
| 4 mixed baseline | 2.443 | 2.449 | 2.473 | 1847 s | 2.217 |
| 4 mixed two-tier source policy | 1.571 | 1.435 | 1.486 | 2402 s | 1.705 |
| **4 mixed two-tier + owner hint** | **1.531** | **1.406** | **1.436** | **2422 s** | **1.691** |

### 2P2D

2P2D baseline, two-tier source policy, and owner-hint redundancy comparison

### 4 mixed

4 mixed baseline, two-tier source policy, and owner-hint redundancy comparison

Charts and private benchmark/plotting helpers are not committed to the PR.

## Validation

- `cargo test -p pegaflow-metaserver` (32 passed)
- `cargo test -p pegaflow-core --no-default-features --features cuda-13,rdma` (134 passed, 1 ignored; integration and doc tests passed)
- `cargo clippy -p pegaflow-metaserver --all-targets -- -D warnings`
- `cargo clippy -p pegaflow-core --all-targets --no-default-features --features cuda-13,rdma -- -D warnings`
- `cargo check -p pegaflow-server --no-default-features --features cuda-13,rdma`
- release workspace build and `maturin develop --release`
- latest 2P6D E2E: 2048/2048 requests completed
- previous 2P2D and 4-mixed E2E: 4096/4096 requests completed in each topology

The E2E host runs CUDA 12.8, so runtime binaries/native bindings used CUDA 12. Rust test and clippy validation also covered the CUDA 13 + RDMA feature set.
## Review follow-up (`dc26c86`)

- Reclaim hints now count post-insert **visible/live owners**, matching `query_prefix` and redundancy metrics. Stale, superseded, and missing-node owner records do not satisfy the `>= 3` threshold.
- A new session at an existing node address is treated as a new owner session; repeated inserts from the same session remain idempotent and do not emit duplicate hints. Added regression coverage for stale owners and session takeover.
- Removed the four single-call Insert helpers left behind after `InsertWithHint` was removed. Request construction, response handling, and `FailedPrecondition` session reset now follow the existing Remove loop shape without changing queue-drop, failure-count, client-reset, or flush-ack behavior.
- Kept the internal `MetaServerClient(config, Weak<ReadCache>)` construction path and did not add an unused compatibility wrapper or change the RPC contract.

### Review validation

- `cargo fmt --all -- --check` passed.
- `cargo test -p pegaflow-metaserver` passed: 34 tests.
- `cargo clippy -p pegaflow-metaserver --all-targets -- -D warnings` passed.
- Core CUDA/RDMA test and clippy commands were attempted locally but are blocked by the macOS environment: `rdma-mummy-sys` requires Linux headers, and `io-uring` requires Linux APIs. The existing Linux CI jobs remain the validation path for those feature combinations.

## Review follow-up (88dfc7f)

Addressed the three batch-path nits without changing hint semantics:

- Skip the `Weak<ReadCache>` upgrade and cache lock when an Insert response contains no `reclaimable_hashes`; the batch API also short-circuits empty input.
- Aggregate resident-class gauge updates by the number of blocks that actually moved, producing at most one decrement and one increment per batch. Missing, already-reclaimable, duplicate, and cross-namespace hashes remain no-ops.
- Stop the visible-owner scan once the `>= 3` threshold is reached; stale/superseded/session handling is unchanged.

Validation on the remote Linux environment:

- `cargo test -p pegaflow-metaserver`: 34 passed
- `cargo test -p pegaflow-core --lib storage::read_cache::tests`: 16 passed
- `cargo clippy -p pegaflow-metaserver --all-targets -- -D warnings`
- `cargo clippy -p pegaflow-core --all-targets -- -D warnings`

### DIFF
--- a/pegaflow-core/src/internode/metaserver_client.rs
+++ b/pegaflow-core/src/internode/metaserver_client.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::sync::Weak;
 
 use log::{debug, error, info, warn};
 use pegaflow_common::grpc::{GRPC_CLIENT_HTTP2_KEEPALIVE_INTERVAL, GRPC_CONNECT_TIMEOUT};
@@ -16,6 +17,7 @@ use tonic::transport::{Channel, Endpoint};
 use uuid::Uuid;
 
 use crate::metrics::core_metrics;
+use crate::storage::ReadCache;
 
 // Shared insert/remove command channel depth. Eviction bursts outrun the single
 // consumer's per-RPC drain, so a shallow queue silently drops removals.
@@ -92,8 +94,6 @@ struct BlockHashBatch {
     groups: Vec<(String, Vec<Vec<u8>>)>,
 }
 
-type InsertHintCallback = Box<dyn FnOnce(Vec<u32>) + Send + 'static>;
-
 impl BlockHashBatch {
     fn from_entries(entries: Vec<(String, Vec<u8>)>) -> Self {
         let mut groups: HashMap<String, Vec<Vec<u8>>> = HashMap::new();
@@ -119,10 +119,6 @@ impl BlockHashBatch {
 /// Command sent to the background MetaServer loop.
 enum MetaServerCommand {
     Insert(BlockHashBatch),
-    InsertWithHint {
-        batch: BlockHashBatch,
-        callback: InsertHintCallback,
-    },
     Remove(BlockHashBatch),
     /// Barrier: acked once every insert/remove enqueued before it has been
     /// delivered to the MetaServer (or dropped after a failed attempt).
@@ -158,7 +154,7 @@ impl MetaServerClient {
     /// Create a new client and spawn the background registration loop.
     ///
     /// Must be called from within a tokio runtime context.
-    pub fn new(config: MetaServerClientConfig) -> Self {
+    pub(crate) fn new(config: MetaServerClientConfig, read_cache: Weak<ReadCache>) -> Self {
         let endpoint = metaserver_endpoint(config.metaserver_addr.clone());
         let (command_tx, rx) = mpsc::channel(config.queue_depth);
 
@@ -167,6 +163,7 @@ impl MetaServerClient {
             config.metaserver_addr.clone(),
             endpoint.clone(),
             config.advertise_addr,
+            read_cache,
         ));
 
         // Lazy-connect query client: connects on first RPC, not here
@@ -197,30 +194,6 @@ impl MetaServerClient {
             return;
         }
         self.try_send_register_batch(BlockHashBatch::single_namespace(namespace, hashes));
-    }
-
-    /// Enqueue a registration whose response carries post-insert owner counts.
-    /// Queue, RPC, or protocol failures leave cache classification unchanged.
-    pub(crate) fn try_register_namespace_with_hint<F>(
-        &self,
-        namespace: String,
-        hashes: Vec<Vec<u8>>,
-        callback: F,
-    ) where
-        F: FnOnce(Vec<u32>) + Send + 'static,
-    {
-        if hashes.is_empty() {
-            return;
-        }
-        let batch = BlockHashBatch::single_namespace(namespace, hashes);
-        let count = batch.count();
-        self.try_send_registration(
-            MetaServerCommand::InsertWithHint {
-                batch,
-                callback: Box::new(callback),
-            },
-            count,
-        );
     }
 
     fn try_send_register_batch(&self, batch: BlockHashBatch) {
@@ -367,6 +340,7 @@ async fn registration_loop(
     metaserver_addr: String,
     endpoint: Endpoint,
     advertise_addr: String,
+    read_cache: Weak<ReadCache>,
 ) {
     let mut client: Option<MetaServerGrpcClient<Channel>> = None;
     let node_id = Uuid::new_v4().to_string();
@@ -417,7 +391,6 @@ async fn registration_loop(
         let mut inserts: HashMap<String, Vec<Vec<u8>>> = HashMap::new();
         let mut removes: HashMap<String, Vec<Vec<u8>>> = HashMap::new();
         let mut mixed_ops: Option<HashMap<(String, Vec<u8>), bool>> = None; // true=insert
-        let mut hinted_inserts: Vec<(BlockHashBatch, InsertHintCallback)> = Vec::new();
         let mut saw_insert = false;
         let mut saw_remove = false;
         // Flush barriers drained in this batch; acked after the sends below, so
@@ -438,16 +411,6 @@ async fn registration_loop(
                     } else {
                         append_groups(&mut inserts, batch.groups);
                     }
-                }
-                MetaServerCommand::InsertWithHint { batch, callback } => {
-                    saw_insert = true;
-                    if saw_remove {
-                        let net = mixed_ops.get_or_insert_with(|| {
-                            build_net(std::mem::take(&mut inserts), std::mem::take(&mut removes))
-                        });
-                        clear_pending_removes(net, &batch.groups);
-                    }
-                    hinted_inserts.push((batch, callback));
                 }
                 MetaServerCommand::Remove(batch) => {
                     saw_remove = true;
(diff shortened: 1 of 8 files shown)


pegaflow-core/src/internode/metaserver_client.rs
```diff
@@ -487,9 +488,15 @@ async fn registration_loop(
                 match c.insert_block_hashes(request).await {
                     Ok(resp) => {
                         let inner = resp.into_inner();
+                        if let Some(cache) = read_cache.upgrade() {
+                            cache.mark_reclaimable_hashes(namespace, &inner.reclaimable_hashes);
```
xiaguan, line 494: Nit: could we skip the `ReadCache` upgrade/lock when `reclaimable_hashes` is empty? Most inserts will not return hints.

pegaflow-core/src/storage/read_cache.rs
```diff
@@ -149,6 +160,25 @@ impl ReadCache {
             .add(-retained_blocks, &*CACHE_CLASS_RETAINED);
         removed
     }
+
+    pub(crate) fn mark_reclaimable_hashes(&self, namespace: &str, hashes: &[Vec<u8>]) {
+        let mut inner = self.inner.lock();
+        for hash in hashes {
+            let key = BlockKey::new(namespace.to_string(), hash.clone());
+            mark_reclaimable(&mut inner, &key);
+        }
```
xiaguan, line 185: Nit: since this is already a batch API, would you mind short-circuiting on an empty `hashes` slice and aggregating the resident-class metric updates once per batch instead of per key?

pegaflow-metaserver/src/store.rs
```diff
@@ -205,21 +207,33 @@ impl BlockHashStore {
...
                     key_register_time: now,
                 },
             );
+            let is_new_owner = previous.is_none_or(|owner| owner.node_id != node_id);
+            if is_new_owner
+                && owners
+                    .iter()
+                    .filter(|(node, owner)| self.is_owner_visible(node, owner, now))
+                    .count()
+                    >= MIN_RECLAIMABLE_OWNER_COUNT
```
xiaguan: Nit: for the `>= 3` check, could we stop counting visible owners once we reach the threshold rather than scanning the full owner set each time?

xiaguan (commented):
Thanks for this change — the MetaServer hint path and live-owner threshold look solid overall, and the 2P6D redundancy results are convincing. A few small batch-path nits below; none block the core design.

pegaflow-core/src/internode/metaserver_client.rs
GentleCold: Addressed in 88dfc7f: the registration loop now upgrades ReadCache only when reclaimable_hashes is non-empty; the batch API also short-circuits empty input.

pegaflow-core/src/storage/read_cache.rs
GentleCold: Addressed in 88dfc7f: the batch tracks only retained-to-reclaimable transitions and applies one aggregated decrement/increment pair per non-empty batch. Missing, duplicate, and already-reclaimable hashes remain no-ops.

pegaflow-metaserver/src/store.rs
GentleCold: Addressed in 88dfc7f: visible-owner counting now uses take(MIN_RECLAIMABLE_OWNER_COUNT), so it stops after the threshold while preserving the existing live-owner filter and >= 3 behavior.

xiaguan (approved):
LGTM.

The MetaServer live-owner reclaim hints, local demote path, and the 88dfc7f batch-path follow-ups look good. Insert cost is bounded (especially after `take(3)`), and query contention remains the existing per-shard / sync-handler shape rather than a new global stall. Fine to merge.

PR #414, opened 2026-07-22
feat(server): native VMM KV clients over an fd side-channel

xiaguan:
## Summary

Adds a **native (non-Python) KV registration path** that shares GPU KV memory with the engine by **VMM POSIX file descriptor** instead of a CUDA IPC handle.

Why: a CUDA-IPC-imported pointer **cannot be registered into the NIC** — `ibv_reg_mr` on a `cuIpcOpenMemHandle` pointer fails, and no dma-buf fd can be obtained for it. It can therefore never back GPUDirect RDMA. A VMM allocation (`cuMemCreate` + POSIX-fd export) **can** be handed to `ibv_reg_dmabuf_mr`. This PR is the foundation for RDMA-capable KV offload (the RDMA registration itself is a follow-up).

A POSIX fd cannot travel inside a gRPC message (it is a host-local kernel object), so the client sends it over a **host-local Unix-domain socket via `SCM_RIGHTS`**, keyed by `(instance_id, device_id)`, before `register_context_batch`. The server advertises the socket path in `HealthResponse`.

## Changes

- **proto**: replace `CudaIpcTensor` with `NativeKvTensor` (the fd travels out of band, not in the message); add `HealthResponse.fd_socket_path` and `RegisterContextRequest.native_alloc_size`. Version tag → `+native-vmm-v1`.
- **`fd_channel.rs`**: Unix-socket listener receiving one fd per connection via `recvmsg`/`SCM_RIGHTS`, delivered to the matching registration with a short wait for reordering.
- **`registry.rs`**: `VmmMapping` imports the fd (`cuMemImportFromShareableHandle` + reserve/map/set-access) and tears it down in order; one shared allocation per batch, each layer a strided view into it.
- **`--python-registry`** (default `true`): keeps the torch/vLLM connector path. `false` runs a **torch-free** deployment that initializes device contexts via cudarc and serves only native clients.

## Testing

- Native **Qwen3-4B host-tier save/load verified end to end**: a prefix evicted from HBM is restored from the CPU tier **bit-exact** (head-logprob delta 0.0000). Verified in both `--python-registry true` and torch-free modes.
- Server unit tests: 23 passed.
- Consumer vs data-center GPU: `cuMemCreate` rejects `gpuDirectRDMACapable` on parts that don't support it, so the client sets that flag only where the device attribute reports support.

## Not in this PR

- NIC registration (`ibv_reg_dmabuf_mr`) of the imported allocation — the mapping is RDMA-ready but wiring RDMA is a follow-up.
- Multi-arena (GLM5.2) VMM export; this PR covers the single fused Qwen3 buffer.

### DIFF
--- a/pegaflow-server/src/fd_channel.rs
+++ b/pegaflow-server/src/fd_channel.rs
@@ -76,11 +76,6 @@ impl FdChannel {
         Ok(channel)
     }
 
-    /// Absolute socket path to advertise to clients (via `HealthResponse`).
-    pub fn path(&self) -> &str {
-        &self.path
-    }
-
     /// Claim the fd registered for `key`, waiting up to `timeout` for it to
     /// arrive if the registration RPC beat its fd. Returns `None` on timeout.
     pub async fn take(&self, key: FdKey, timeout: Duration) -> Option<OwnedFd> {
--- a/pegaflow-server/src/fd_channel.rs
+++ b/pegaflow-server/src/fd_channel.rs
@@ -166,7 +166,9 @@ fn recv_fd_with_key(sock: RawFd) -> std::io::Result<(FdKey, OwnedFd)> {
     // SAFETY: cmsg points into our control buffer.
     let (level, ctype) = unsafe { ((*cmsg).cmsg_level, (*cmsg).cmsg_type) };
     if level != libc::SOL_SOCKET || ctype != libc::SCM_RIGHTS {
-        return Err(std::io::Error::other("fd side-channel: unexpected control message"));
+        return Err(std::io::Error::other(
+            "fd side-channel: unexpected control message",
+        ));
     }
     // SAFETY: CMSG_DATA points at the fd payload of a SCM_RIGHTS message.
     let raw_fd = unsafe { std::ptr::read_unaligned(libc::CMSG_DATA(cmsg).cast::<RawFd>()) };
--- a/pegaflow-server/src/fd_channel.rs
+++ b/pegaflow-server/src/fd_channel.rs
@@ -16,33 +16,40 @@ pub(crate) struct RegistrationKey {
     pub(crate) device_id: i32,
 }
 
-/// Bound UDS path plus fds waiting to be claimed by `register_context_batch`.
-#[derive(Clone)]
-pub struct FdChannel {
+struct Inner {
     path: String,
-    pending: Arc<Mutex<HashMap<RegistrationKey, OwnedFd>>>,
-    arrived: Arc<Notify>,
+    pending: Mutex<HashMap<RegistrationKey, OwnedFd>>,
+    arrived: Notify,
 }
+
+impl Drop for Inner {
+    fn drop(&mut self) {
+        remove_stale_socket(&self.path);
+    }
+}
+
+/// Bound UDS path plus fds waiting to be claimed by `register_context_batch`.
+/// Cloning only bumps the `Arc`; the socket file is removed when the last clone drops.
+#[derive(Clone)]
+pub struct FdChannel(Arc<Inner>);
 
 impl FdChannel {
     pub fn bind(path: String) -> std::io::Result<Self> {
         remove_stale_socket(&path);
         let listener = UnixListener::bind(&path)?;
-        let channel = Self {
+        let channel = Self(Arc::new(Inner {
             path,
-            pending: Arc::new(Mutex::new(HashMap::new())),
-            arrived: Arc::new(Notify::new()),
-        };
-        let pending = Arc::clone(&channel.pending);
-        let arrived = Arc::clone(&channel.arrived);
+            pending: Mutex::new(HashMap::new()),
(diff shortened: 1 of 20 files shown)


xiaguan:
Re-trigger CI after fix(ci) push

pegaflow-server/src/fd_channel.rs
```diff
@@ -0,0 +1,364 @@
...
+/// `instance_id\0device_id` payload; generous so long instance ids do not truncate.
+const MAX_KEY_PAYLOAD: usize = 4096;
+/// Control buffer for one SCM_RIGHTS fd.
+const CMSG_SPACE_FOR_FD: usize = unsafe {
+    // SAFETY: CMSG_SPACE is a pure layout calculation from the payload length.
+    libc::CMSG_SPACE(std::mem::size_of::<RawFd>() as libc::c_uint) as usize
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct RegistrationKey {
```
GentleCold, line 23: [blocking] Could this key include a per-registration nonce (and have the same nonce in the gRPC request)? The current `(instance_id, device_id)` key lets a late fd from a timed-out registration remain pending and be consumed by a retry or a new process using the same instance id. If the old and new allocations have the same layout, the import succeeds while save/load silently targets the wrong GPU buffer. Pending fds also need TTL/capacity cleanup.

pegaflow-server/src/registry.rs
```diff
@@ -11,22 +13,119 @@ pub struct TensorMetadata {
...
+}
+
+impl Drop for VmmMapping {
+    fn drop(&mut self) {
+        self.context
+            .bind_to_thread()
+            .expect("bind CUDA context before releasing VMM mapping");
+        // SAFETY: reverse of import: unmap, free VA, release handle.
+        unsafe {
+            sys::cuMemUnmap(self.base_ptr, self.size_bytes)
```
GentleCold, line 113: [blocking] This unmaps the VMM allocation as soon as the registry owner is dropped, but the engine can still have asynchronous load/save work running on detached GPU worker threads. `unregister` and session cleanup drop the registry before removing the engine instance, so a worker can access `base_ptr` after `cuMemUnmap`. Please make the mapping lifetime follow the transfer tasks, or stop/drain/join the workers before releasing it.

pegaflow-server/src/fd_channel.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub(crate) async fn take(&self, key: RegistrationKey, timeout: Duration) -> Option<OwnedFd> {
+        let deadline = tokio::time::Instant::now() + timeout;
+        loop {
+            {
+                let mut pending = self.0.pending.lock().expect("fd pending map poisoned");
+                if let Some(fd) = pending.remove(&key) {
+                    return Some(fd);
+                }
+            }
+            if tokio::time::timeout_at(deadline, self.0.arrived.notified())
```
GentleCold, line 85: [blocking] There is a lost-wakeup window here: `take` checks `pending`, then creates `notified()`, while `recv_one` can insert the fd and call `notify_waiters()` between those operations. `notify_waiters` does not leave a permit for a future created afterward, so the fd can be present while this RPC sleeps until the 60s timeout. Register and enable the waiter before checking the map, or use a per-key channel.

pegaflow-server/src/lib.rs
```diff
@@ -463,17 +489,26 @@ pub fn run() -> Result<(), Box<dyn Error>> {
         return Err("No CUDA devices available".into());
     }
 
-    init_python_cuda(&devices)?;
+    if cli.python_registry {
+        init_python_cuda(&devices)?;
+    } else {
+        init_cudarc_cuda(&devices)?;
```
GentleCold, line 495: [blocking] The advertised torch-free mode still imports torch when `--devices` is omitted: automatic device detection above calls `detect_cuda_devices()`, which imports `torch` before this branch is reached. A native-only server with no torch therefore cannot use the default auto-detection path. Device discovery should use CUDA Driver/cudarc when `python_registry` is false.

pegaflow-server/src/fd_channel.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
+
+/// UDS side-channel; last `Arc` drop unlinks the socket file.
+#[derive(Clone)]
+pub struct FdChannel(Arc<Inner>);
+
+impl FdChannel {
+    pub fn bind(path: String) -> std::io::Result<Self> {
+        remove_stale_socket(&path);
```
GentleCold, line 46: [P2] `remove_stale_socket` unconditionally unlinks the configured path before bind. With the default `/tmp/pegaflow-fd.sock`, starting a second server can disconnect the first server, and the first server may later unlink the second server's socket on drop. Please use an instance-specific path or only remove a path after proving the previous listener is stale/owned by this process.

PR #418, opened 2026-07-30
fix(connector): hold blocks until all save tasks finish

xiaguan:
## Summary

- track pending save tasks per request instead of treating pending work as a boolean
- report `finished_sending` only after every queued task for the request completes
- clear stale completion state when a later save task is enqueued

## Failure mechanism

A request can enqueue multiple `SaveTask` instances across scheduler steps. The old set-based state marked the whole request complete after the first task. vLLM could then release and reuse GPU blocks while a later task still referenced their block IDs, allowing bytes from another request to be stored under the original block hash.

The regression test models that reuse directly. Against the old implementation, the value stored under `second hash` is `reused by another request`; with this change it remains `second request block`.

## Validation

- regression tests against old implementation: 2 failed as expected
- regression tests with this change: 2 passed
- source-only default Python gate: 204 passed, 11 deselected
- pre-commit hooks: passed, including release Cargo tests and Python lint/format checks
- H200 x8 GLM-5.2 FP8 clean round-trip: 606 external blocks hit, 38,784 cached prompt tokens, no replacement characters or repeated-exclamation corruption
- pre-fix synthetic GPU stress did not reproduce the production gibberish; this PR remains draft until the post-fix model-specific E2E is complete

## Scope

This proves and fixes a KV save lifecycle corruption path. It does not yet claim that this was the only cause of the observed production gibberish.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,8 @@ jobs:
           persist-credentials: false
       - name: Run ruff check
         uses: astral-sh/ruff-action@v3
+        with:
+          version: "0.14.14"
 
   python-unit-tests:
     name: python unit tests
--- a/python/pegaflow/connector/worker.py
+++ b/python/pegaflow/connector/worker.py
@@ -177,7 +177,7 @@ class WorkerConnector:
         )
         self._save_thread.start()
 
-        self._req_pending_saves: set[str] = set()
+        self._req_pending_save_tasks: dict[str, int] = {}
         self._completed_saves: set[str] = set()
         self._save_completion_lock = threading.Lock()
         self._save_completion_events: dict[str, threading.Event] = {}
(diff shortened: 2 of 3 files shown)


GentleCold:
Non-blocking test coverage suggestion: please add a `handle_preemptions()` regression test where the same request has two pending save tasks. The test should verify that preemption does not return until both tasks complete, since this hook is the final guard before vLLM can reuse the request's blocks. I did not identify this as a confirmed bug in the current implementation; adding the test would make the concurrency fix evidence more complete.

xiaguan:
Addressed in e527929. The new regression test enqueues two save tasks for the same request, runs `handle_preemptions()` on a separate thread, waits until it has entered the completion-event wait, and verifies that the first task does not release preemption while the second task does.

Validation:
- `origin/master`: fails because preemption returns after the first task
- this branch: targeted test passes (3/3 lifecycle tests)
- default Python gate: 205 passed
- source-only Python gate: 205 passed

Thanks for calling out this final block-reuse guard.

GentleCold (approved):
LGTM

PR #419, opened 2026-07-31
feat(metrics): add cache residence TTL dashboard

GentleCold:
## Summary

- record RAM cache residence duration from first insertion to removal as the `pegaflow_cache_residence_duration_seconds` histogram
- distinguish pressure eviction from cleanup removal with a fixed low-cardinality `reason` label and 1-second to 24-hour buckets
- add Grafana p50/p95/p99 and heatmap panels, a removal-reason selector, and metrics documentation

### Rust and dashboard checks

- remote H200: `cargo fmt` check passed
- remote H200: `pegaflow-core` clippy passed with CUDA 13, RDMA, and warnings denied
- remote H200: library tests passed (144 passed, 1 ignored)
- remote H200: `engine_basic` GPU roundtrip tests passed (9 passed)
- remote H200: explicit CUDA kernel/direct test passed (1 passed)
- remote H200: remaining non-SSD integration tests passed (12 passed)
- local: dashboard JSON parsing, panel ID uniqueness, grid overlap, selector/PromQL consistency, and `git diff --check` passed
- CI: all checks passed, including Rust fmt/clippy, CUDA 12/13 cargo checks, Python unit tests, Ruff, and CUDA 12/13 wheel builds

### RTX 4090 vLLM 2P2D integration test

Environment:

| Item | Configuration |
| --- | --- |
| GPU host | 8 x RTX 4090 (`192.168.172.117`) |
| PegaFlow | 4 instances (2 prefill + 2 decode), 100 GB pool per instance |
| vLLM | 2P2D service with proxy and four engines |
| Model | `/data/models/DeepSeek-V2-Lite-Chat` |
| Transfer | NIXL with CPU buffer |
| Observability | Prometheus continuously scraped all four PegaFlow targets; Grafana imported the Direct Prometheus dashboard |

Multiturn workload:

- 256 conversations, 32 concurrent clients, 8 turns per conversation
- approximately 8,192 input tokens per turn and 128 output tokens
- 1,024 requests over 899.105 seconds
- 1.139 requests/second with zero request failures

Observed after the run:

| Metric | Value |
| --- | ---: |
| TTL / eviction observations | 438,784 |
| Residence duration sum | 169,195,707.67 s |
| Resident cache bytes | 91,574,157,312 |
| Reclaimed bytes | 1,788,841,558,016 |

The Grafana view below uses `reason=pressure` and the exact active-data window, `2026-07-31 20:27:30` to `20:43:00 CST`. The initial p95/p99 values near two hours are expected: blocks left resident by the earlier warmup/load were evicted when this sustained run applied pressure. The steady-state quantiles then fall to the seconds/minutes range.

PegaFlow cache residence TTL Grafana dashboard after the RTX 4090 multiturn run

## Known validation limit

The full `pegaflow-core` run has 11 existing `ssd_cache` failures on the H200 host. Its 8 GPU-local NUMA nodes split the fixed 32 KiB test pool into 4 KiB per node, below the 16 KiB test batch; serial execution reproduces the same topology-specific failure. The TTL implementation does not touch the SSD cache path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,8 @@ jobs:
           persist-credentials: false
       - name: Run ruff check
         uses: astral-sh/ruff-action@v3
+        with:
+          version: "0.14.14"
 
   python-unit-tests:
     name: python unit tests
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -102,6 +102,27 @@ PegaFlow exposes the following metrics for monitoring KV cache operations:
   - Current sealed block bytes resident in cache (sum of footprints)
   - Use case: Attribute pinned pool usage to cache residency
 
+- **pegaflow_cache_residence_duration_seconds** (Histogram)
+  - RAM resident block lifetime from its first successful cache insertion to
+    removal, measured in seconds
+  - Labels: `reason` (`pressure` or `cleanup`)
+  - `pressure`: allocator pressure removed the block through LRU reclaim
+  - `cleanup`: the memory-cache cleanup endpoint removed the block
+  - Buckets: `1s`, `5s`, `10s`, `30s`, `1m`, `2m`, `5m`, `10m`, `30m`,
+    `1h`, `2h`, `6h`, `12h`, `24h`, and `+Inf`
+  - Use case: Track typical and tail cache residence time and visualize the
+    eviction-age distribution
+  - Scope: each RAM residence is a separate lifetime. A block reinserted after
+    eviction starts a new lifetime; cache hits, duplicate inserts, and
+    replacement-class changes do not reset the original insertion time.
+  - The lifetime ends when the block leaves the resident cache, even if an
+    outstanding `Arc` keeps its pinned memory allocated. Correlate
+    `pegaflow_cache_block_evictions_still_referenced_total` and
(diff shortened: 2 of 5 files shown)


xiaguan (approved):
Reviewed the metric wiring end to end:

- Removal-path coverage: TinyLfuCache is unbounded (no internal eviction), and the only two exits (reclaim loop -> pressure, cleanup endpoint -> cleanup) both flow through record_residence_durations, so every RAM residence episode is observed exactly once.
- Lifetime semantics match the docs: duplicate inserts and cache hits preserve the original insertion time, class migration carries the metadata, reinsert after eviction starts a new episode (all covered by unit tests).
- Recording happens outside the cache mutex, so no critical-section growth.
- Exported name pegaflow_cache_residence_duration_seconds follows the existing unit-suffix convention (cf. pegaflow_save_duration_seconds); docs, PromQL examples, and dashboard selectors all line up. Dashboard panel IDs unique, no grid overlap.

Pushed 252daf4 for two review nits: align the CoreMetrics field name with the instrument name, and make RemovedResident::inserted_at non-optional so the type expresses the resident-block metadata invariant (debug_assert + removal-time fallback kept for release builds).

Local: pegaflow-core lib tests 144 passed; pre-commit hooks (fmt/clippy/release tests) green.

LGTM, will auto-merge once the wheel builds finish.

PR #421, opened 2026-08-04
feat(connector): support vLLM hybrid KV cache models

xiaguan:
## Summary

- extend the load RPC with cache-grouped layers and explicit per-group destination presence
- support FullAttention + Mamba hybrid cache models in `align` mode across registration, save, query, and load
- preserve real TP rank identity for hybrid MLA while retaining pure-MLA TP deduplication
- source recurrent saves from committed vLLM cache mappings and load only the final recurrent state
- reject unsupported sparse-attention and non-align Mamba combinations at startup
- prevent direct DMA coalescing across distinct host or device allocations

## Breaking change

The load RPC now accepts grouped layers and per-group destination targets. Compatibility with older clients and servers is intentionally not preserved.

## Validation

- Python default gate: 224 passed, 12 deselected
- Python source-only gate: 224 passed, 12 deselected
- Cargo workspace tests with CUDA 13 and RDMA features
- full pre-commit gate: format, clippy, release tests, Ruff, and typos passed
- CPython 3.12 CUDA 13 + RDMA wheel built and installed successfully
- Qwen3.5-4B vLLM correctness E2E: 5 passed
- Kimi-K3 hybrid-cache wheel E2E on B300 × 8: 5/5 exact cold/warm outputs at 0/1536/3072/4608/6144 cached-token boundaries
- five-sample A/B at 4608 cached tokens: 1.1215s cold median, 0.4632s warm median, 2.42× median speedup, 5/5 exact outputs
- cross-process vLLM restart E2E: 4608 cached tokens, exact output, 1.0019s cold to 0.3487s warm
- toxic-reviewer final verdict: Ready for Review

### DIFF
--- a/python/pegaflow/connector/__init__.py
+++ b/python/pegaflow/connector/__init__.py
@@ -12,6 +12,7 @@ import torch
 from vllm.distributed.kv_transfer.kv_connector.v1.base import (
     KVConnectorBase_V1,
     KVConnectorRole,
+    SupportsHMA,
 )
 from vllm.distributed.parallel_state import get_pp_group, get_tensor_model_parallel_rank
 
@@ -33,7 +34,7 @@ from pegaflow.connector.worker import WorkerConnector
 from pegaflow.pegaflow import EngineRpcClient
 
 
-class PegaKVConnector(KVConnectorBase_V1):
+class PegaKVConnector(KVConnectorBase_V1, SupportsHMA):
     """v1 KV connector for PegaFlow with separated scheduler/worker logic."""
 
     def __init__(self, vllm_config, role: KVConnectorRole, kv_cache_config=None):
@@ -143,7 +144,8 @@ class PegaKVConnector(KVConnectorBase_V1):
         # is silently ignored upstream and falls back to per-layer — surface
         # that instead of pretending the request was honored.
         env_cross_layer = os.environ.get("PEGAFLOW_CROSS_LAYER_BLOCKS", "1") == "1"
-        self._prefer_cross_layer = env_cross_layer and not is_mla
+        cache_groups = tuple(getattr(kv_cache_config, "kv_cache_groups", ()) or ())
+        self._prefer_cross_layer = env_cross_layer and not is_mla and len(cache_groups) <= 1
         if is_mla and env_cross_layer:
             logger.warning(
                 "[PegaKVConnector] PEGAFLOW_CROSS_LAYER_BLOCKS=1 is ignored for MLA "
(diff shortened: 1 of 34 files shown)


python/pegaflow/connector/common.py
```diff
@@ -115,10 +118,96 @@ class LoadIntent:
...
+                layer_names=((),),
+                hash_group_index=0,
+                has_recurrent_state=False,
+                recurrent_group_indices=frozenset(),
+                recurrent_layer_names=frozenset(),
+            )
+
+        from vllm.v1.kv_cache_interface import FullAttentionSpec, MambaSpec
+
+        if len(groups) > 1 and any(
```
GentleCold: Guarding these checks with `len(groups) > 1` misses single-cache-group, non-`FullAttentionSpec` HMA configurations. vLLM considers HMA required whenever any group is not `FullAttentionSpec`; the existing NIXL implementation also explicitly handles the single-group SW-only case. As written, a single `SlidingWindowSpec` or other sparse spec, as well as a single Mamba group with `mamba_cache_mode != "align"`, is silently accepted and then enters an unsupported block/hash mapping path. Please validate the spec type and Mamba mode for every group, reject HMA configurations without a dense FullAttention hash group, and add coverage for single-group sparse attention and non-align Mamba.

python/pegaflow/connector/__init__.py
```diff
@@ -43,13 +44,15 @@ def __init__(self, vllm_config, role: KVConnectorRole, kv_cache_config=None):
         tp_size = vllm_config.parallel_config.tensor_parallel_size
         world_size = vllm_config.parallel_config.world_size
         is_mla = detect_mla(vllm_config)
+        cache_groups = tuple(getattr(kv_cache_config, "kv_cache_groups", ()) or ())
```
GentleCold, line 47: `cache_groups` is available here, but it is not included in `derive_namespace()`. Running the same model once with the default HMA configuration and once with `--disable-hybrid-kv-cache-manager` therefore produces the same namespace and block hashes even though the registered layer/slot layouts differ. With a shared server or during a rolling configuration change, the query can hit blocks written by the other layout and fail during load with `stored block has N slots but instance expects M`. Please include a stable cache-group signature in the namespace (at minimum the HMA setting, spec types, block size/Mamba mode, and group layer layout) and add a test showing that HMA-enabled and HMA-disabled configurations use different namespaces.

GentleCold (commented):
Static review found two HMA edge cases; see the inline comments for details.

GentleCold (approved):
LGTM

PR #422, opened 2026-08-05
fix(transfer): chunk pega_batch_copy launches at 64MB

xiaguan:
## Problem

`KernelBackend::submit` issues one `pega_batch_copy` launch for a whole restore batch — one CTA per ~34KB fragment, grid up to 65535. A multi-GB restore parks CTAs across every SM for **470–550ms**, and CUDA never preempts running CTAs, so co-resident decode kernels starve for the full duration. On an EP fleet this is amplified: one starved rank stalls every peer inside the all-to-all dispatch, so a single restore turns into a fleet-wide latency spike.

Caught by dual-node nsys on a GLM5.2 2P+2D agent-trace replay (GB300, EP8 decode): during each stall window a single `pega_batch_copy` occupied the culprit GPU at ~100% while decode kernels totalled 0.1–0.4ms; culprit GPUs rotate with restore traffic.

## Fix

Split `submit` into 64MB chunks (one launch each, ~2000 CTAs). Each launch bounds the SM monopoly window to ~10–20ms; kernels queued on other streams interleave between chunks. Scratch capacity is ensured for the largest chunk **before** the first launch — growing mid-loop would free the descriptor buffer an in-flight chunk still reads.

## Validation (same workload/config A/B, GLM5.2 replay, kernel transfer mode)

| metric | before | after |
|---|---|---|
| ITL p99 | 706 ms | **156 ms** |
| TPOT p99 | 273 ms | 84 ms |
| output throughput | 713 tok/s | 873 tok/s (+22%) |
| copy-kernel p50 / p99 | one 470–550ms launch per restore | 0.34 ms / 23 ms |
| fleet-wide stall events | continuous (~18/min) | cold-start bursts only |

Residual 300–825ms outliers in the profile are chunks stretched by SM *sharing* with decode — decode keeps running through them, which is the intended trade.

### DIFF
--- a/pegaflow-core/src/transfer/kernel.rs
+++ b/pegaflow-core/src/transfer/kernel.rs
@@ -47,6 +47,13 @@ extern "C" __global__ void pega_batch_copy(const unsigned long long* __restrict_
 
 const BLOCK_DIM: u32 = 256;
 const MAX_GRID: u32 = 65535;
+/// Upper bound on bytes moved per kernel launch. A single launch covering a
+/// whole multi-GB restore parks one CTA per fragment across every SM for
+/// hundreds of ms; CUDA never preempts running CTAs, so co-resident decode
+/// kernels starve for the full duration — and on an EP fleet one starved rank
+/// stalls every peer's dispatch. Chunked launches bound that monopoly window
+/// to ~10-20ms; queued kernels from other streams interleave between chunks.
+const CHUNK_BYTES: usize = 64 << 20;
 
 /// Single-launch transfer backend. Compiled once per worker; the descriptor
 /// scratch buffer is reused across calls.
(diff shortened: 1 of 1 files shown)


xiaguan:
Closing in favor of a simpler endgame on the engine side: we're moving the GLM5.2 KV cache to a GPU page-first slab and registering it with pegaflow as a **single arena** (segment = whole per-block page, block_stride = page stride). That makes every restore one contiguous copy per block in Direct mode — no batch-copy kernel on the SM at all, so there's nothing left to chunk. Until that lands our stack runs from a patched checkout carrying this fix. If other layer-first users hit the same SM-monopoly stall, this patch is still the right shape — happy to reopen.

PR #423, opened 2026-08-07
feat(connector): support host-local shards for cross-host TP

xiaguan:
## Summary

- add ordered `pegaflow.tp_shard_endpoints` configuration for one TP replica spanning multiple hosts
- remap global TP ranks to host-local server topology and isolate each shard namespace
- query every shard, use the common prefix, and carry one exact lease per shard to workers
- open one session watcher per shard and reject unsupported PP, DCP, PCP, uneven, or duplicate layouts
- add unit coverage plus a focused multi-server vLLM E2E gate

## Supported topology

Endpoints map equal contiguous global TP rank ranges. For TP8 with two endpoints, ranks 0-3 use shard 0 and ranks 4-7 use shard 1. Each server only opens CUDA IPC handles for GPUs on its own host.

## Validation

- `uv run ruff check .`
- default Python gate: 251 passed
- source-only Python gate: 251 passed
- targeted connector tests: 98 passed
- `cargo test --release`
- staged `prek run`
- toxic reviewer: READY
- vLLM E2E on 4 GPUs: TP4 split into two local TP2 servers, Qwen3-4B, kernel transfer backend, 96 cold plus 96 warm requests at concurrency 8; outputs matched, both shards recorded save/load activity, and all RPCs succeeded

Measured request traffic:

- cold: 33.81 req/s, p50 224 ms, p95 321 ms
- warm: 42.74 req/s, p50 187 ms, p95 203 ms

## Pressure note

A separate undersized pinned-pool experiment triggered eviction, but also exposed existing `pinned pool exhausted` save failures under forced pressure. This PR does not claim eviction-pressure robustness and keeps that allocator behavior out of scope.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,7 +1896,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pegaflow-common"
-version = "0.23.6"
+version = "0.23.7"
 dependencies = [
  "colored",
  "libc",
@@ -1906,7 +1906,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-core"
-version = "0.23.6"
+version = "0.23.7"
 dependencies = [
  "ahash",
  "bytesize",
@@ -1939,7 +1939,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-metaserver"
-version = "0.23.6"
+version = "0.23.7"
 dependencies = [
  "axum",
  "clap",
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-pd-wire"
-version = "0.23.6"
+version = "0.23.7"
 dependencies = [
  "serde",
  "serde_json",
@@ -1967,7 +1967,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-proto"
-version = "0.23.6"
+version = "0.23.7"
 dependencies = [
  "prost",
  "tonic",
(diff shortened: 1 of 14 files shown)


xiaguan:
Two-container TP4 validation completed on a single 8-GPU development host.\n\nTopology:\n- One TP4 vLLM replica on four GPUs.\n- Two isolated PegaFlow containers, each serving one contiguous TP2 shard.\n- Shared CUDA IPC namespace; separate gRPC endpoints.\n- Qwen3-4B, max model length 4096.\n\nCorrectness and warm-cache traffic (direct backend):\n- 96 cold + 96 warm requests, concurrency 8.\n- Cold: 33.45 req/s, p50 224 ms, p95 312 ms.\n- Warm after vLLM restart: 39.36 req/s, p50 193 ms, p95 225 ms.\n- Cold and warm outputs matched exactly.\n- Both shards recorded save and load activity.\n- RPC failures: 0 on both shards.\n\nCache-pressure follow-up:\n- 96 additional long requests with independent first blocks, concurrency 4.\n- All 96 requests completed with non-empty outputs.\n- Shard 0: 14,998 evictions, 18.61 GB reclaimed.\n- Shard 1: 15,254 evictions, 24.27 GB reclaimed.\n- RPC failures remained 0 on both shards.\n\nA separate container-only issue was observed with the kernel transfer backend: context registration succeeds, but the first descriptor scratch allocation returns CUDA_ERROR_OUT_OF_MEMORY even with about 19 GiB free per GPU. The same PR passed the non-container two-server TP4 kernel-backend run, so this appears independent of TP-shard endpoint routing and needs separate CUDA/container investigation.\n\nAll PR CI checks are green.

feifei-111 (approved):

PR #426, opened 2026-08-11
fix(connector): support uniformly grouped MLA cache specs

xiaguan:
## Summary

- accept a single UniformTypeKVCacheSpecs group only when it is non-empty and every inner spec is exactly MLAAttentionSpec
- use the single logical group as the block-hash group while preserving all MLA and sparse-indexer layer names
- keep the existing multi-group FullAttention plus aligned Mamba capability boundary unchanged
- cover GLM-style MLA plus indexer layouts and rejection of mixed or empty wrappers

## Validation

- default Python gate: 255 passed, 13 deselected
- source-only Python gate: 255 passed, 13 deselected
- pre-commit hooks, including cargo test --release
- constructed the layout with real vLLM main MLAAttentionSpec and UniformTypeKVCacheSpecs objects using different main-attention and indexer head sizes

### DIFF
--- a/python/pegaflow/connector/common.py
+++ b/python/pegaflow/connector/common.py
@@ -260,13 +260,24 @@ class CacheGroupLayout:
             FullAttentionSpec,
             MambaSpec,
             MLAAttentionSpec,
+            UniformTypeKVCacheSpecs,
         )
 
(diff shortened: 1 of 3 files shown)


GentleCold (approved):
LGTM

PR #432, opened 2026-08-13
fix(connector): discard drifted TP prefetch results

xiaguan:
## Summary

- keep polling an in-flight prefetch with its original query identity when local prefix hits change
- let TP shard validation check the stale result against the query that created it
- release stale per-shard leases before issuing the updated query

## Root cause

The scheduler already snapshots `computed_blocks` and query hashes when `query_prefetch` returns Loading. After TP shard support was added, each shard validated a later Ready result against the current, shorter query before the scheduler could compare it with the saved snapshot. A valid stale result therefore raised `hits > queried blocks` instead of being discarded.

## Tests

- `cd python && uv run --extra test pytest -q` (257 passed, 13 deselected)
- `cd python && uv run --extra test pytest -q tests/test_tp_shards.py tests/test_combine_hashes.py` (66 passed)
- commit hooks, including `cargo test --release`, ruff, format, typos, and Commitizen

### DIFF
--- a/python/pegaflow/connector/scheduler.py
+++ b/python/pegaflow/connector/scheduler.py
@@ -237,9 +237,16 @@ class SchedulerConnector:
             self._release_pending_query_probe(req_id)
             probe = None
 
-        # No reusable Ready result.  Ask backend.
+        # A Loading task is keyed by req_id server-side. If the current query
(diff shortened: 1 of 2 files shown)


GentleCold (approved):
LGTM

PR #434, opened 2026-08-17
fix(core): keep leased blocks indexed under pressure

xiaguan:
## Summary

- reclaim only cache entries exclusively owned by the resident cache
- keep leased, queried, loading, and weakly referenced blocks indexed during memory pressure
- preserve reclaimable-before-retained ordering with a bounded LRU scan
- strengthen the GPU eviction integration test to prove an unleased batch is reclaimed while a leased batch remains queryable and loadable

## Testing

- `cargo test --no-default-features --features cuda-13,rdma -p pegaflow-core storage::read_cache::tests -- --nocapture`
- `cargo test --no-default-features --features cuda-13,rdma -p pegaflow-core --test eviction -- --nocapture`
- `prek run` (includes release-mode Rust test suite)

### DIFF
--- a/pegaflow-core/src/cache.rs
+++ b/pegaflow-core/src/cache.rs
@@ -72,6 +72,16 @@ impl TinyLfuCache<BlockKey, ArcSealedBlock> {
         self.lru.contains_key(key)
     }
 
+    /// Returns true only when the cache owns the block exclusively.
+    ///
+    /// `Arc::get_mut` rejects both other strong references and weak references,
+    /// so no holder outside the cache can keep or reacquire the allocation.
+    pub(crate) fn is_exclusively_owned(&mut self, key: &BlockKey) -> bool {
+        self.lru
+            .peek_mut(key)
+            .is_some_and(|block| Arc::get_mut(block).is_some())
+    }
+
     /// Insert with TinyLFU admission. If the candidate is colder than the
     /// current LRU victim it is dropped.
     pub(crate) fn insert(&mut self, key: BlockKey, value: ArcSealedBlock) -> CacheInsertOutcome {
--- a/pegaflow-core/tests/eviction.rs
+++ b/pegaflow-core/tests/eviction.rs
@@ -11,7 +11,7 @@ use pegaflow_core::StorageConfig;
 
 const BLOCK_SIZE: usize = 4096;
 const NUM_BLOCKS: usize = 4;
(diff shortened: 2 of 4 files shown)


Kevin-XiongC (approved):

pegaflow-core/src/cache.rs
```diff
@@ -72,6 +72,16 @@ impl TinyLfuCache<BlockKey, ArcSealedBlock> {
         self.lru.contains_key(key)
     }
 
+    /// Returns true only when the cache owns the block exclusively.
+    ///
+    /// `Arc::get_mut` rejects both other strong references and weak references,
+    /// so no holder outside the cache can keep or reacquire the allocation.
+    pub(crate) fn is_exclusively_owned(&mut self, key: &BlockKey) -> bool {
```
GentleCold, line 79: [P1] `Arc::get_mut` also returns `None` when the allocation has only a `Weak` reference. The SSD write path intentionally converts every pending block to `Weak<SealedBlock>` before enqueueing it (`write_path.rs`), and the writer upgrades it only after dequeuing. While that queue is backed up, this new check treats those blocks as pinned, so `remove_lru_batch()` can return empty even though there is no external strong owner. With a full pool, the next save then fails allocation instead of reclaiming the block. This conflicts with the existing SSD contract that weak refs must not prevent eviction. Please distinguish lease/load strong references from transient SSD weak refs, for example by checking strong ownership only or using an explicit pin state.

pegaflow-core/tests/eviction.rs
```diff
@@ -11,7 +11,7 @@ use pegaflow_core::StorageConfig;
 
 const BLOCK_SIZE: usize = 4096;
 const NUM_BLOCKS: usize = 4;
-/// Pool fits one batch with headroom for metadata, but not two.
+/// Pool fits two batches, so a third batch must reclaim one of them.
```
GentleCold, line 14: [P2] This pool-size change removes eviction pressure from two existing tests. `eviction_reclaims_old_blocks_for_new` and `eviction_works_with_sharded_pool` still save only two batches, while the pool now intentionally fits two batches. Neither test asserts that the first batch is gone, so both can pass even if eviction is broken. Please keep a third batch in these tests or add an explicit `count_hits_then_release(old)` assertion so the original eviction and sharded-eviction coverage remains meaningful.

PR #436, opened 2026-08-17
fix(core): allow SSD weak refs during pressure eviction

GentleCold:
## Summary

- Treat only additional strong `Arc<SealedBlock>` owners as pressure-eviction pins.
- Allow SSD write-queue `Weak<SealedBlock>` references to yield to memory pressure, preserving the existing fire-and-forget SSD contract.
- Add regression coverage for weak-only eviction and strong-reference retention.

## Validation

Ran on RTX 4090 / Linux with CUDA 13 and RDMA enabled:

- `cargo test --no-default-features --features cuda-13,rdma -p pegaflow-core storage::read_cache::tests -- --nocapture` (22 passed)
- `cargo test --no-default-features --features cuda-13,rdma -p pegaflow-core --test eviction -- --nocapture --test-threads=1` (3 passed)

The existing `ssd_cache` integration suite was also attempted on the 4090 host, but its fixed 32 KiB test pool is split across the host's NUMA-local pools and exhausts before the first save; all 11 failures are `pinned pool exhausted`, before SSD assertions run.

### DIFF
--- a/pegaflow-core/src/cache.rs
+++ b/pegaflow-core/src/cache.rs
@@ -72,14 +72,14 @@ impl TinyLfuCache<BlockKey, ArcSealedBlock> {
         self.lru.contains_key(key)
     }
 
-    /// Returns true only when the cache owns the block exclusively.
+    /// Returns true when the cache is the only strong owner of the block.
     ///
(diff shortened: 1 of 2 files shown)


xiaguan (approved):
LGTM

PR #438, opened 2026-08-17
feat(metaserver): aggregate HLL reports across nodes

GentleCold:
## Summary
- send mergeable raw HLL window snapshots with node heartbeats
- union active node registers in MetaServer and export cluster HLL gauges
- keep HLL reporting best-effort; metadata liveness is not affected by invalid reports or metadata writes
- preserve the existing miss-only local HLL semantics and default windows

## Validation
- `cargo test -p pegaflow-common` (51 passed)
- `cargo test -p pegaflow-metaserver` (42 passed)
- `cargo test -p pegaflow-core --no-default-features --features cuda-13,rdma internode::metaserver_client::tests` (8 passed)
- `cargo clippy -p pegaflow-common -p pegaflow-metaserver --all-targets -- -D warnings`
- `cargo clippy -p pegaflow-core -p pegaflow-server --no-default-features --features cuda-13,rdma --all-targets -- -D warnings`
- `cargo check -p pegaflow-core -p pegaflow-server --no-default-features --features cuda-13,rdma`

Full core unit tests were also attempted; CUDA-dependent tests cannot run on the remote environment because `libcudart` is unavailable.

### DIFF
--- a/pegaflow-metaserver/src/metric.rs
+++ b/pegaflow-metaserver/src/metric.rs
@@ -86,9 +86,8 @@ pub fn register_store_gauges(store: &Arc<BlockHashStore>) {
                 observer.observe(avg, &[]);
             })
             .build();
-        // Each HLL gauge re-merges the cluster union in its own callback; at
-        // scrape frequency the merge cost (nodes x windows x registers) is
-        // negligible, and OTEL 0.31 no longer offers multi-instrument callbacks.
+        // The store cache makes these callbacks share one cluster union while
+        // preserving separate instruments in OpenTelemetry 0.31.
         let hll_cardinality_store = Arc::clone(&s);
         let hll_cardinality = meter
             .f64_observable_gauge("pegaflow_metaserver_hll_cardinality")
--- a/pegaflow-metaserver/src/metric.rs
+++ b/pegaflow-metaserver/src/metric.rs
@@ -16,6 +16,11 @@ struct StoreGaugeHandles {
     _nodes: ObservableGauge<u64>,
     _redundancy: ObservableGauge<u64>,
     _redundancy_avg: ObservableGauge<f64>,
+    _hll_cardinality: ObservableGauge<f64>,
+    _hll_total_requests: ObservableGauge<u64>,
+    _hll_estimated_hit_rate: ObservableGauge<f64>,
+    _hll_active_nodes: ObservableGauge<u64>,
+    _hll_snapshot_age: ObservableGauge<f64>,
 }
 
 static STORE_GAUGES: OnceLock<StoreGaugeHandles> = OnceLock::new();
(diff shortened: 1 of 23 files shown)


pegaflow-metaserver/src/store.rs
```diff
@@ -143,16 +171,38 @@ impl BlockHashStore {
...
 
-    pub fn heartbeat_node(&self, node: &str, node_id: Uuid) -> Result<(), StoreError> {
+    pub fn heartbeat_node(
+        &self,
+        node: &str,
+        node_id: Uuid,
+        hll_report: HllNodeReport,
+    ) -> Result<(), StoreError> {
         let now = Instant::now();
+        let validated_report = match validate_hll_report(&hll_report) {
```
xiaguan, line 181: Could we make schema selection and report insertion atomic? `hll_schema_matches_active()` scans the map before the later `nodes.entry(...)` update, so concurrent initial heartbeats can both observe no active schema and store incompatible reports. I reproduced this after prior reports expired: 64 new sessions remained active, but only 9 reports were included because conflicting schemas were admitted and subsequent reports rejected one another. The aggregation path also `zip`s register arrays, which silently truncates if incompatible lengths reach it. Since the HLL schema is deployment configuration, one simpler option would be to configure the expected schema in MetaServer and reject a mismatch with `FailedPrecondition` during startup/registration. That removes first-writer ordering and the check-then-insert race entirely.

pegaflow-metaserver/src/store.rs
```diff
@@ -398,7 +482,7 @@ impl BlockHashStore {
     }
 
     fn touch_node_session(&self, node: &str, node_id: Uuid) -> Result<(), StoreError> {
```
xiaguan, line 484: This changes metadata liveness semantics in a way that can hide a healthy node. The client drains and awaits all insert/remove chunks before returning to its heartbeat `select`; if one drain lasts longer than `node_stale_after`, these successful metadata writes no longer refresh `last_seen`, and `query_prefix()` filters every owner from this node. The HLL report already has its own receipt timestamp and `cluster_hll_snapshot()` checks it independently, so restoring the existing `last_seen` refresh here would keep metadata availability without extending a stale HLL report.

pegaflow-metaserver/src/store.rs
```diff
@@ -454,6 +601,74 @@ impl BlockHashStore {
...
+            ));
+        }
+        payload_bytes = payload_bytes.saturating_add(window.registers.len());
+        if !(MIN_BUCKET_BITS..=MAX_BUCKET_BITS).contains(&window.bucket_bits) {
+            return Err(format!(
+                "HLL bucket_bits must be in {MIN_BUCKET_BITS}..={MAX_BUCKET_BITS}, got {}",
+                window.bucket_bits
+            ));
+        }
+        let expected_registers = 1usize << window.bucket_bits;
```
xiaguan, line 638: The length check does not validate register values. A report with the correct length, `total_requests = 1`, and every register set to 255 is accepted; in a release-mode probe it produced a cardinality of `6.234246084537104e77` and contaminated the union until expiry. Because the client hashes `(namespace, block_hash)` to exactly 8 bytes, the valid maximum register value is derivable from `bucket_bits`. Could this invariant live in a checked `HyperLogLog::try_from_registers` constructor so raw register mutation cannot bypass it?

pegaflow-metaserver/src/metric.rs
```diff
@@ -81,12 +86,88 @@ pub fn register_store_gauges(store: &Arc<BlockHashStore>) {
                 observer.observe(avg, &[]);
             })
             .build();
+        // Each HLL gauge re-merges the cluster union in its own callback; at
+        // scrape frequency the merge cost (nodes x windows x registers) is
+        // negligible, and OTEL 0.31 no longer offers multi-instrument callbacks.
+        let hll_cardinality_store = Arc::clone(&s);
```
xiaguan, line 92: Each of the five callbacks clones all active reports and recomputes the complete register union. With the default three 65,536-register windows, I measured one snapshot at about 19.66 ms for 64 nodes and the five callbacks at about 96.80 ms per collection. The independent calls can also observe different heartbeat generations, so cardinality, total requests, and hit rate may not describe the same aggregate. Could we cache one aggregate per report generation/collection interval and have all five instruments read that snapshot?

xiaguan (changes requested):
Thanks for putting this together. The register-union approach is a good fit for cluster-level HLL aggregation, and the affected test suites pass locally. I found a few same-version correctness and concurrency issues that I think should be addressed before merging; rolling-upgrade and compatibility concerns are intentionally out of scope here.

Performance data collected on the same host with release builds:
- Existing `unregister_node` benchmark: master medians 720.71/697.15 ms; PR medians 647.16/689.75 ms. The second-run intervals overlap, so this does not establish a performance change.
- New aggregation path with the default three windows: 1/8/64 nodes took 0.85/2.75/19.66 ms for one snapshot, while the five current gauge callbacks took 4.21/13.84/96.80 ms per collection.

There is also a module-boundary concern: `pegaflow-common/src/hll.rs` grows from 961 to 1,020 lines, and `pegaflow-metaserver/src/store.rs` grows from 1,037 to 1,395 lines. Extracting report validation/schema ownership/aggregation into a dedicated MetaServer HLL module would make the invariants easier to enforce and keep the block metadata store focused.

Finally, the four commits currently do not contain `Signed-off-by` trailers; please add them if DCO is required for this repository.

pegaflow-metaserver/src/store.rs
GentleCold: Fixed in 7c35f86. Valid insert/remove calls now refresh NodeRecord.last_seen after session validation, preserving metadata owner visibility. HLL receipt time remains stored separately in ReceivedHllReport and is refreshed only by a valid heartbeat report, so metadata activity cannot keep an expired report in the cluster union. Added a regression that asserts both behaviors together.

pegaflow-metaserver/src/metric.rs
GentleCold: Fixed in 7c35f86. All five gauges now read a shared generation-aware aggregate cache; report changes are coalesced for 250 ms so one collection stays on one snapshot, while the earliest report expiry still invalidates immediately. Accepted reports are Arc-backed, and union work happens only on cache refresh. Release medians for 64 nodes/default windows are 0.720 ms for one snapshot and 0.720 ms for five gauge-style reads, versus the 19.66/96.80 ms baseline. Cache reuse, generation coalescing, and expiry have regression tests.

pegaflow-metaserver/src/store.rs
GentleCold: Fixed in 7c35f86. MetaServer now constructs one immutable expected HLL schema from the shared startup windows/bucket-bits config and validates every report against it in exact order, so there is no first-writer or check-then-insert race. The union also rejects unequal register lengths explicitly. HLL mismatch remains best-effort and does not fail heartbeat/session liveness. Added concurrent incompatible-initial-report and unequal-length regression tests.

pegaflow-metaserver/src/store.rs
GentleCold: Fixed in 7c35f86. Report validation now rejects any register above 65 - bucket_bits before it can enter node state or the union; a 255-register regression test covers the contamination case. I kept this check at the MetaServer report boundary because the generic HyperLogLog accepts hashes longer than eight bytes, while this wire report specifically comes from the stable 64-bit namespaced identity.

PR #439, opened 2026-08-19
fix(core): bound SSD prefetch staging allocations

xiaguan:
## Summary

- Replace one whole-batch SSD prefetch slab per NUMA node with bounded 256 MiB staging chunks.
- Preserve per-slot NUMA placement, slot ordering, oversized-slot handling, and the existing all-or-nothing batch allocation contract.
- Add a representative multi-group regression shape that verifies the allocation bound.

## Production evidence before this change

Observed on one GB300 4-GPU node running a hybrid multi-group workload:

| Metric | Observed value |
| --- | ---: |
| SSD-prefetched data | 83 blocks / 8.52 GB |
| Largest contiguous staging request | 3.5 GiB per NUMA node |
| Cache blocks evicted | 4,608 |
| Logical cache bytes evicted | 415.6 GiB |
| Allocator bytes reclaimed | 462.8 GB |
| Pinned-pool usage | 484.5 GB -> 30.2 GB |

The allocator reclaims until its largest free extent can satisfy the request. A multi-GiB whole-batch slab therefore turns fragmentation into a large eviction burst. This is the SSD equivalent of the RDMA staging issue addressed by #387.

## Allocation-shape A/B

The deterministic multi-group regression uses 56 blocks with eight 16 MiB slots split evenly across two NUMA nodes:

| Allocation shape | Baseline | This PR |
| --- | ---: | ---: |
| Bytes staged per NUMA | 3.5 GiB | 3.5 GiB |
| Maximum contiguous allocation | 3.5 GiB | 256 MiB |
| Allocations per NUMA | 1 | 14 |
| Slots preserved | 224 | 224 |

This establishes the allocation bound. Runtime eviction and hit-ratio impact still need to be re-measured after deployment.

## Validation

- `cargo clippy -p pegaflow-core --all-targets --no-default-features --features cuda-13,rdma -- -D warnings`
- `cargo test -p pegaflow-core --no-default-features --features cuda-13,rdma -- --test-threads=1` (154 passed, 1 existing GPU-only test ignored)
- `prek run`

### DIFF
--- a/pegaflow-core/src/backing/ssd_cache.rs
+++ b/pegaflow-core/src/backing/ssd_cache.rs
@@ -712,8 +712,8 @@ struct SlotPlacement {
 }
 
 struct PrefetchChunk {
+    capacity: u64,
     size: u64,
-    used: u64,
     slots: Vec<SlotPlacement>,
 }
 
@@ -757,9 +757,9 @@ fn chunk_slot_refs(refs: &[SlotRef], chunk_bytes: u64) -> Result<Vec<PrefetchChu
     for slot in refs {
         let needs_new_chunk = current.as_ref().is_none_or(|chunk| {
             chunk
-                .used
+                .size
                 .checked_add(slot.size)
-                .is_none_or(|end| end > chunk.size)
+                .is_none_or(|end| end > chunk.capacity)
         });
 
         if needs_new_chunk {
@@ -767,8 +767,8 @@ fn chunk_slot_refs(refs: &[SlotRef], chunk_bytes: u64) -> Result<Vec<PrefetchChu
                 chunks.push(chunk);
             }
             current = Some(PrefetchChunk {
-                size: remaining.min(chunk_bytes).max(slot.size),
-                used: 0,
+                capacity: remaining.min(chunk_bytes).max(slot.size),
+                size: 0,
                 slots: Vec::new(),
             });
         }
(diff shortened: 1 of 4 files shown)


pegaflow-core/src/backing/ssd_cache.rs
```diff
@@ -739,37 +808,43 @@ async fn dispatch_prefetch_batch(
...
-
-        let mut offset = 0usize;
-        for r in refs {
-            slot_allocs[r.block_idx][r.slot_idx] = Some(SlotAlloc {
-                allocation: allocation.clone(),
-                offset,
-            });
-            offset += r.size as usize;
+        for chunk in chunks {
+            let allocation = match store.allocate_prefetch(chunk.size, *numa_node) {
```
GentleCold, line 829: [P1] Allocate the bytes actually used by each planned chunk

`chunk.size` is the packing capacity, but it can be larger than `chunk.used` when the next slot does not fit in the current chunk. Allocating `chunk.size` here therefore retains the unused tail of every non-full chunk.

For example, with a 256 MiB cap and slots of `[200, 100, 100] MiB`, the planner produces:

```text
(size=256, used=200)
(size=200, used=200)
```

The logical batch is 400 MiB, but this path allocates 456 MiB. A single-shard pool with exactly 400 MiB available could satisfy the old whole-batch allocation, while the new path fails on the second chunk and returns an empty batch, potentially after reclaiming the resident cache for the first allocation. This changes the allocation-success envelope rather than only bounding contiguous allocation size.

Please allocate `chunk.used` here (or shrink each finalized chunk size to its used bytes), and add a non-divisible-slot regression asserting that the sum of allocation sizes equals the sum of slot sizes. The current 16 MiB shape divides 256 MiB exactly, so it cannot catch this case.

GentleCold (approved):
LGTM

PR #441, opened 2026-08-26
fix(connector): store HMA recurrent checkpoint once at request end

xiaguan:
## Summary
- Hybrid (HMA) requests now skip mid-flight saves and emit a single recurrent checkpoint from `request_finished`.
- That checkpoint is resolved through vLLM's committed prefix cache instead of the live align-mode block table, which otherwise stores a mid-block running state or a speculative draft block under the wrong hash.
- Regression tests build the recurrent table with vLLM's align-mode formulas and encode provenance in block ids so a wrong save is visible.

This is not duplicating an existing PR: the 5090 exploration session never landed a commit or PR.

## Test plan
- [x] `cd python && uv run --extra test pytest` — 278 passed, 13 deselected
- [x] `tests/test_hma_checkpoint_index.py` — 9 cases (mid-block / aligned / speculative, plus skip-when-stored and new-boundary-only)
- [x] Local Qwen3.5-4B GPU invariant script (`logs/k3-hma/verify_hma_invariants.py`): vbs=528, mid-block keys last committed boundary, short skip, warm hit no re-save with matching greedy text, prefix-extend files only the new boundary, independent miss, two-block checkpoint at 1056
- [x] `pytest -m e2e tests/test_vllm_e2e_correctness.py --model /data/models/Qwen3.5-4B --max-model-len 4096` — 5 passed (9m30s); same-process HMA load used PegaFlow; no data-path RPC / KV load failures

### DIFF
--- a/python/pegaflow/connector/scheduler.py
+++ b/python/pegaflow/connector/scheduler.py
@@ -521,11 +521,7 @@ class SchedulerConnector:
             # is reset on preemption — no connector-side bookkeeping can be
             # trusted across a preempt/resume cycle).
             written = req.num_computed_tokens + num_tokens
-            if save_intent := self._consume_save_intent(
-                req_id,
-                written,
-                req.num_computed_tokens,
-            ):
+            if save_intent := self._consume_save_intent(req_id, written):
                 potential_saves[req_id] = save_intent
 
         # Process cached (running) requests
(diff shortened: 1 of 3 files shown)


GentleCold (approved):
LGTM

PR #444, opened 2026-08-28
Restore reconciled HMA local prefix cache hits

xiaguan:
## Summary

- replace the HMA-wide local cache suppression with a scoped PegaFlow adapter
- reuse only the prefix already reconciled across attention and recurrent cache groups
- leave unbound vLLM block pools unchanged and restore the original coordinator method on shutdown
- fail early if another runtime patch changes the adapted vLLM lookup while PegaFlow is active
- extend the correctness E2E to verify that same-process HMA hits stay local while cross-process hits still load through PegaFlow

No vLLM source changes are required.

## Why

The previous connector replaced `BlockPool.get_cached_block` with a function that always returned `None` for HMA layouts. That avoided exposing a dense attention prefix before recurrent state was usable, but it also disabled every valid L1 prefix hit.

vLLM already exposes a reconciled hybrid lookup through `HybridKVCacheCoordinator.find_longest_cache_hit`. The adapter uses that result for PegaFlow-bound block pools and reports the same safe boundary for every cache group.

## Measured A/B

Qwen3.5-4B, one GPU, 3,168 cached prompt tokens, one output token, seven exact warm requests:

| Revision | Warm median | PegaFlow hits | PegaFlow load bytes | Matching outputs |
| master | 101.90 ms | 42 blocks | 1,089,994,752 | 7/7 |
| this PR | 77.09 ms | 0 blocks | 0 | 7/7 |

Warm request latency decreased by 24.35% in this workload.

## Validation

- `cd python && uv run --extra test pytest`: 272 passed, 13 deselected
- source-only Python gate: 272 passed, 13 deselected
- Qwen3.5 correctness E2E: 5 passed
  - same-process HMA request reported a local cached-token hit with zero PegaFlow hit/load delta
  - cross-process phase reported 40 saved blocks / 692.1 MB and 10 hit blocks / 432.5 MB
  - output equality, RPC failure, and KV load failure checks passed
- `prek run`: passed, including `cargo test --release`, ruff, formatting, and typo checks

Example E2E command:

```bash
cd python
uv run --extra test pytest -m e2e tests/test_vllm_e2e_correctness.py \
  --model /path/to/Qwen3.5-4B --max-model-len 4096 --pegaflow-pool-size 2gb
```

### DIFF
--- a/python/pegaflow/connector/hma_local_cache.py
+++ b/python/pegaflow/connector/hma_local_cache.py
@@ -1,0 +1,101 @@
+"""Scoped vLLM adapter for coherent HMA local prefix hits."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from threading import Lock
+from typing import Any
+
+_lock = Lock()
+_bound_pools: dict[int, tuple[object, int]] = {}
+_coordinator_type: type | None = None
+_original_lookup: Callable[..., Any] | None = None
+_adapted_lookup: Callable[..., Any] | None = None
+
+
+def enable_reconciled_hma_local_hits(block_pool: object) -> None:
+    """Make vLLM report only locally resumable hybrid prefixes for this pool."""
+    global _adapted_lookup, _coordinator_type, _original_lookup
+
+    with _lock:
+        pool_id = id(block_pool)
+        existing = _bound_pools.get(pool_id)
+        if existing is not None:
+            if existing[0] is not block_pool:
+                raise RuntimeError("HMA local-cache block-pool identity collision")
+            _bound_pools[pool_id] = (block_pool, existing[1] + 1)
+            return
+
+        if _coordinator_type is None:
+            from vllm.v1.core.kv_cache_coordinator import HybridKVCacheCoordinator
+
+            coordinator_type = HybridKVCacheCoordinator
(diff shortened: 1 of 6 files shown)


xiaguan:
Closing in favour of #448 (released as v0.24.1).

The hit-rate problem on hybrid models turned out to be checkpoint placement, not local-cache suppression: the connector stored one recurrent checkpoint per request at `request_finished`, so any shared prefix that did not end exactly there could not be resumed. #448 saves every boundary state vLLM hands off (`kv_connector_block_state.boundary_state_offloads`) and hints the junction via `Request.shared_prefix_boundary`, which fixes the cross-session / branching case this PR did not.

What remains of this PR is an L1 latency optimisation: with HMA every warm request loads through PegaFlow instead of hitting vLLM's GPU prefix cache. In our P/D deployment that is a few ms of same-node IPC copy per 10k-token hit (0.8 ms per load in the #448 E2E), sessions are not pinned to one P node anyway, and an L1 hit needs the same local recurrent checkpoint an external hit does. Not worth carrying the vLLM coordinator adapter for right now. If profiling ever shows PegaFlow load time as a visible cost, this can be revisited on top of master (it will need a rebase over the #448 changes to `scheduler.py` and re-validation with boundary-job pinning).

PR #445, opened 2026-08-28
feat(connector): cache linear state on GPU

UranusSeven:
The idea is to create a small mamba state pool backed by HBM so that pegaflow can focus on MLA cache.

The pool should work as long as the number of slots is larger than the number of on-going sessions, which is less than 100 for a prefill node so that the total size of the pool is around 50GB.


UranusSeven:
#441 should do the job.

PR #448, opened 2026-09-02
fix(connector): save every HMA recurrent boundary state vLLM hands off

xiaguan:
## Problem

On hybrid models (Kimi-K3 / Kimi-Linear: KDA recurrent state + MLA attention) the PegaFlow connector's hit rate on agent workloads is far below the Mooncake store connector's, and it is not a capacity issue. The connector stored **one** recurrent checkpoint per request, at `request_finished`, taken from whatever boundary block vLLM still had cached. Agent traffic shares a system prompt across sessions and branches mid-history, so the shared prefix almost never ends at that single checkpoint. `reconcile_hybrid_hit` then finds the attention pages but no usable recurrent state and logs

```
[PegaKVConnector] req=... HMA attention prefix of N blocks has no common recurrent checkpoint; recomputing instead
```

i.e. a 0-block hit on exactly the traffic Mooncake serves from the second session on. (#444 does not address this; it only restores local hits.)

## Fix

Store recurrent states the way vLLM offers them (same hand-off the Mooncake connector consumes since vllm-project/vllm#51358):

- **Per-boundary saves.** Consume `SchedulerOutput.kv_connector_block_state.boundary_state_offloads` every step. Each `(group, block, boundary_tokens)` entry becomes a boundary save job filed under the request hash ending at that boundary: chunk ends, internal prefill checkpoints, every block crossed while decoding. The block is pinned in the GPU block pool (`pool.touch`) until every worker reports the job via the new `PegaWorkerMetadata` (`KVConnectorWorkerMetadata.aggregate` across TP), so the save is decoupled from the request's block lifetime; `has_pending_push_work` keeps the engine stepping while jobs are in flight. HMA now requires a vLLM with this hand-off and fails loudly without it.
- **Attention pages save mid-flight** at the normal per-block cadence with recurrent rows nulled; recurrent blocks are never saved positionally from the connector's table mirror. Request saves run asynchronously again (the synchronous D2H in `wait_for_save` is gone).
- **Junction hint.** When the external attention prefix runs past the last usable checkpoint, set `Request.shared_prefix_boundary` so vLLM stops the prefill chunk there and commits (and hands off) the state at the end of the shared prefix. The first sharer recomputes once; every later sharer resumes.

The first commit is a prerequisite for running on current vLLM main at all: vLLM now allocates one KV buffer and hands each layer a strided `[B, H, N, C]` view at a nonzero storage offset, which registration rejected (zero-offset assertion; MLA split ratio read from the head axis).

## Verification

Kimi-Linear-48B-A3B-Instruct, single GB300, vLLM main (2026-09-01 nightly, 1920-token blocks, mamba `align`), `pegaflow-server` 40 GiB pool, greedy decoding. Prompts share a ~10.5k-token system prefix `S` (5 full blocks); E/F share only the first ~4.3k tokens of `S` (2 full blocks) and then diverge for ~3k unique tokens, so their common prefix ends where no earlier checkpoint exists.

| request | master | this PR |
|---|---|---|
| A cold | 0 / 5 blocks | 0 / 5 |
| A' exact repeat | 5 | 5 |
| B, C sharers of full `S` | 5 | 5 |
| D next turn of A | 5 | 5 |
| E first sharer of the 2-block prefix | 0 (`no common recurrent checkpoint`) | 0 (`no common recurrent checkpoint`, junction committed) |
| **F second sharer of the 2-block prefix** | **0** (`no common recurrent checkpoint`) | **2 / 2 blocks** |

Greedy output of A' equals A in both runs; no load/save failures; `pegaflow_cache_block_hits_total` matches the per-request hit counts.

Unit gate: `uv run --isolated --no-project --with pytest --with numpy --with 'requests>=2.26.0' pytest` → 288 passed. New `tests/test_hma_boundary_offloads.py` covers pinned saves and release across workers, worker-meta aggregation, unusable/duplicate hand-offs, attention-only mid-request saves, and the junction hint.

https://claude.ai/code/session_016pSeF3SWkaeomAW4GQ7M6z


wz1qqx (approved):
LGTM！

PR #449, opened 2026-09-04
fix(connector): keep full-block save intents inside the resumed block table

xiaguan:
## Summary

Production K3 (TP8, MTP, PegaFlow master with #448) hit this on every attention layer once KV pressure started preempting requests:

```
RuntimeError: save block/hash count mismatch for language_model.model.layers.3.self_attn: blocks=1 hashes=10
```

- **Root cause** (`scheduler.py::_consume_full_block_saves`): the save window was bounded by `base_block_idx + len(table)`, but the mirrored table already contains the externally loaded prefix, so the external offset was counted twice. It stayed masked while the scheduled-token accumulator only covered the current life. After a preemption vLLM resets `num_computed_tokens` and rebuilds the table from the resume chunk, while the accumulator and the first life's offset stay: the window runs past the rebuilt table, Python slicing silently shortens the block IDs, and the prompt hashes (known up front) are not shortened. In save-only mode the same stale watermark could save a partially recomputed block under a valid hash.
- **Fix**: bound the window by the table itself; on resume restart the per-request bookkeeping from the refreshed external hit (`_rebase_resumed_request`), keeping already-stored progress.
- **Blast radius fix** (`worker.py`): the raise killed the save thread, so nothing drained the queue afterwards, every later request stayed held for a save that never completed, and the engine leaked KV cache until restart (2 running / 18 waiting / 2 tok/s in the incident). The save worker now survives a failing batch, a malformed intent drops only its own request's save, and completion is always reported so blocks get released.
- **Second bug, found while reproducing under preemption on Kimi-Linear** (`scheduler.py::_load_block_ids_by_group`): the load's destination vector was sized by the *usable* hit, but the engine walks it in lock step with the query lease. When the hybrid reconcile fell back to an earlier checkpoint (an exact-repeat prompt whose final token must be recomputed), the load carried one target fewer than the lease and the engine rejected it with `query lease block count 5 does not match destination block count 4`, which the worker escalated to "Service unavailable" and the engine died. The probe now keeps the leased block count and the load pads unwanted leased blocks with `None` targets (already skipped by the engine).
- Bumps the version to 0.24.2.

## Test plan

- [x] `tests/test_save_intent_preempt_resume.py`: the production shape (19-block prompt, 9 loaded, preempted after a sub-block chunk, resumed) plus save-only partial-block and rebase cases. All 5 fail on master, pass here.
- [x] `tests/test_connector_save_lifecycle.py`: malformed intent is skipped and the request still completes; save worker survives a failing batch.
- [x] `tests/test_hma_boundary_offloads.py::test_load_targets_cover_every_leased_block_when_the_hit_shrinks`: fails on master, passes here.
- [x] `cd python && uv run --isolated --no-project --with pytest --with numpy --with 'requests>=2.26.0' pytest` on the connector suites (109 passed), ruff check/format clean.
- [x] Single-GPU Kimi-Linear e2e (tray08, 512 GiB pool, 400-block GPU KV budget to force preemption, 2048-token chunks): master died on the lease/destination mismatch under a warm-prefix + 96 long-decode fillers + long-tail scenario; this branch completes 3 rounds (294 requests, 3908 preemptions, 0 save mismatches, 0 load failures). Multi-turn vllm-bench load (128 conv, 3 turns, 2000+ preemptions) is clean on both. The exact save-mismatch sequence from production was not reproduced e2e on this model (its resume hits cover everything stored, so the rebuilt table is never shorter than the hashes); the unit test pins that shape.

https://claude.ai/code/session_01SXoijrttVSfhrQF6EAq5CC

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,7 +1896,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pegaflow-common"
-version = "0.24.1"
+version = "0.24.2"
 dependencies = [
  "colored",
  "libc",
@@ -1906,7 +1906,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-core"
-version = "0.24.1"
+version = "0.24.2"
 dependencies = [
  "ahash",
  "bytesize",
@@ -1939,7 +1939,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-metaserver"
-version = "0.24.1"
+version = "0.24.2"
 dependencies = [
  "axum",
  "clap",
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-pd-wire"
-version = "0.24.1"
+version = "0.24.2"
 dependencies = [
  "serde",
  "serde_json",
@@ -1967,7 +1967,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-proto"
-version = "0.24.1"
+version = "0.24.2"
 dependencies = [
  "prost",
  "tonic",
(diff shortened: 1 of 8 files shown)


wz1qqx (approved):
LGTM！

PR #450, opened 2026-09-08
fix(connector): key blocks by the hash that closes them (prefix-match-unit / hybrid hash_block_size)

xiaguan:
## Symptom

K3 P/D prefill group (TP8 EP8, hybrid mamba-align + MLA, `--prefix-match-unit 128`, pegaflow 0.24.2): the user asks question A and the model fluently answers an unrelated question B. Hit rate looks healthy, nothing crashes.

## Root cause (commit 1)

vLLM computes `Request.block_hashes` every `hash_block_size` tokens (`resolve_kv_cache_block_sizes`). For a single cache group that is the scheduler block, but for hybrid models it is the GCD of the group block sizes, or the explicit `--prefix-match-unit` (128 tokens in production, against 1536-token scheduler blocks). vLLM maps the fine list onto blocks by taking the last hash inside each block (`BlockHashListWithBlockSize`).

The scheduler connector consumed the list 1-to-1 with block ids, so block `i` (tokens `1536*i .. 1536*(i+1)`) was filed under the hash of the first `128*(i+1)` tokens. Any two sessions sharing that much prefix (an agent system prompt of ~6k tokens covers 48 blocks = 73k tokens) served each other's whole conversation, recurrent checkpoint included, so the output was fluent but about the other session.

Log fingerprint from production:

```
cache_lookup: hit_blocks=87 computed_blocks=0 hit_tokens=133632 num_tokens=159364 ... total_query_hashes=1245
```

1245 = 159364 / 128 (fine hashes were sent as block keys), while 87 × 1536 = 133632.

Fix:
- `ConnectorContext.hash_scale` from vLLM's `resolve_kv_cache_block_sizes`; every consumer of `Request.block_hashes` (prefix query, full-block save intents, HMA boundary-state offloads, P/D tail keys) goes through `_request_block_hashes` = `hashes[scale-1::scale]`, vLLM's own rule.
- `_request_block_hashes` refuses more keys than the request has full blocks, so a finer granularity than the connector assumes fails loudly instead of aliasing.
- `hash_block_size` joins the namespace factors. Entries written under the old keys are unreachable after the upgrade: old block `12i+11` sat under exactly the key new block `i` uses, so a warm store would keep serving the wrong blocks.

## Two more ways stale recurrent state reached the store (commit 2)

Found while auditing the connector against the K3 vLLM fork, both verified in code:

- **Boundaries inside an externally loaded prefix.** vLLM allocates a real recurrent block per externally loaded block (`allocate_external_computed_blocks`), the load fills only the checkpoint, and after the load `_update_waiting_for_remote_kv` → `MambaManager.cache_blocks` still offers every block of the loaded prefix as a boundary state. The connector accepted them, saving n-1 never-written blocks under the prefix's own hashes. Boundaries at or below the externally matched prefix are now skipped.
- **Resumed requests under the V2 model runner** (forced by DSpark) arrive in `scheduled_new_reqs`, not `resumed_req_ids`, so the connector never rebased them: the first life's scheduled-token accumulator let it save a block whose recompute had only reached its first rows. The new-request branch now rebases too (a no-op for a fresh request).
- Namespace also gains `block_size`, `mamba_cache_mode`, `mamba_ssm_cache_dtype`: same hash granularity with a different block span yields identical chained hashes for different blocks, and the server's slot-count guard does not tell them apart.

Not in this PR (follow-up): the worker still raises on a hybrid load failure/timeout, taking the whole prefill group down; the K3 fork can recompute hybrid requests from scratch on load failure (`_update_requests_with_invalid_blocks`), so reporting the blocks via `get_block_ids_with_load_errors` instead would let it recover.

## Verification

- Unit tests: `python/tests/test_hash_block_size_scale.py`, new cases in `test_hma_boundary_offloads.py` and `test_save_intent_preempt_resume.py` (each fails with its fix reverted).
- Single-GPU e2e (Kimi-Linear, 1920-token blocks, `--prefix-match-unit 128`): two sessions sharing only a ~3.4k-token system prompt, each with a different codeword in its own document. master answers session A with session B's codeword (hit 7/7 blocks, `total_query_hashes=105` = 13521/128); with this PR A answers its own codeword and its repeat still gets a full hit (`total_query_hashes=7`). Details in the comment below.

Version bump to 0.24.3.

https://claude.ai/code/session_01D88GoN8BhYDLCj9ALdf6uR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,7 +1896,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pegaflow-common"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "colored",
  "libc",
@@ -1906,7 +1906,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-core"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "ahash",
  "bytesize",
@@ -1939,7 +1939,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-metaserver"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "axum",
  "clap",
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-pd-wire"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "serde",
  "serde_json",
@@ -1967,7 +1967,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-proto"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "prost",
  "tonic",
@@ -1977,7 +1977,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-py"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "log",
  "mea",
@@ -1995,7 +1995,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-server"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "axum",
  "clap",
@@ -2027,7 +2027,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-transfer"
-version = "0.24.2"
+version = "0.24.3"
 dependencies = [
  "anyhow",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.24.2"
+version = "0.24.3"
 edition = "2024"
 license = "Apache-2.0"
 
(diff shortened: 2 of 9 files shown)


xiaguan:
## e2e verification (single GPU, Kimi-Linear-48B-A3B, 1920-token blocks, `--prefix-match-unit 128`)

Scenario: session B = shared system prompt + document containing codeword **PELICAN** + question; session A = same system prompt + a different document containing **WALRUS** + the same question. The two sessions share nothing but the ~3.4k-token system prompt.

| stack | B answer | A answer | A repeat | A lookup |
|---|---|---|---|---|
| master (0.24.2) | PELICAN | **PELICAN** (wrong) | PELICAN | hit 7/7 blocks, `total_query_hashes=105` (= 13521 / 128) |
| this PR | PELICAN | WALRUS | WALRUS | hit 0 → recompute, `total_query_hashes=7`; the repeat then hits 7/7 |

master loads B's entire 13440-token KV into A because A's first 7 fine (128-token) hashes, all inside the shared system prompt, are the keys B's 7 blocks were stored under. With the fix the connector logs `block hashes are computed every 128 tokens; keying PegaFlow blocks (1920 tokens) by every 15-th hash` and keys by the hash closing each block, so A no longer aliases B and A's own repeat still gets a full hit.

jimmy-evo (dismissed):

feifei-111 (approved):

PR #452, opened 2026-09-09
fix(connector): drop the block hash past a popped last token

xiaguan:
## Problem

Since 0.24.3 (#450) the K3 prefill groups restart on a fatal EngineCore error raised from `get_num_new_matched_tokens`:

```
RuntimeError: req chatcmpl-...: 360 block hashes give 30 keys at scale 12 for 29 full blocks;
vLLM's hash granularity is finer than the connector's
```

The prefill node runs `MultiConnector = NixlConnector (producer) + PegaKVConnector`. For hybrid (mamba) models NIXL's `get_num_new_matched_tokens` pops the last prompt token on the P side (`_truncate_mamba_request_for_prefill`, so P computes `h(N-1)` and D recomputes the last token) *after* vLLM has already hashed the full prompt, and it never trims `Request.block_hashes` (same invariant break as vllm-project/vllm#49377; Mooncake has the identical pop). MultiConnector calls NIXL first, so PegaFlow sees `num_tokens = N-1` with hashes for `N` tokens.

With `--prefix-match-unit 128` over 1536-token blocks the extra hash only becomes a key when the prompt is exactly 1536-aligned: `N = 46080 = 30 × 1536` gives 360 hashes → 30 keys against `46079 // 1536 = 29` full blocks, which trips the fail-loud guard added in #450. Every such request kills the group; the group comes back and dies on the next aligned prompt.

## Fix

`_request_block_hashes` drops hashes that reach past `request.num_tokens` before scaling them to per-block keys. Exactly one tail hash is the signature of a popped last token; anything more still means vLLM's hash granularity is finer than the connector's and keeps failing loudly, so the contamination guard from #450 stays in place.

The dropped hash fingerprints a prefix the prefiller will never compute (the last block is partial after the pop), so no key is lost.

Second commit bumps the version to 0.24.4 (`Cargo.toml`, `Cargo.lock`, `python/pyproject.toml`) so the squash merge can be tagged and released directly.

## Tests

- `test_hash_past_a_popped_last_token_is_dropped`: 30-block prompt with 360 hashes and `num_tokens = 30 × 1536 − 1` yields 29 keys; an unaligned prompt is unaffected.
- `python/tests`: 304 passed, ruff check/format clean.

https://claude.ai/code/session_0157jRXpY76iVctr9ERAtiLw

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,7 +1896,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pegaflow-common"
-version = "0.24.3"
+version = "0.24.4"
 dependencies = [
  "colored",
  "libc",
@@ -1906,7 +1906,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-core"
-version = "0.24.3"
+version = "0.24.4"
 dependencies = [
  "ahash",
  "bytesize",
@@ -1939,7 +1939,7 @@ dependencies = [
 
 [[package]]
 name = "pegaflow-metaserver"
-version = "0.24.3"
+version = "0.24.4"
 dependencies = [
  "axum",
  "clap",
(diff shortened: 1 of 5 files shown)


wz1qqx (approved):
LGTM！

PR #457, opened 2026-09-18
feat(connector): support GLM-5.3-Flash kpool-tail scratch cache groups

Kevin-XiongC:
## Summary

Makes `PegaKVConnector` work with **GLM-5.3-Flash** (hybrid: sparse-MLA attention + KDA/Mamba recurrent layers + MTP), which previously failed connector init.

GLM-5.3-Flash produces 5 KV cache groups: `UniformTypeKVCacheSpecs(MLAAttentionSpec)` (8960-token blocks), `UniformTypeKVCacheSpecs(KpoolTailSpec)` (4-token blocks, 1 circular block per request), and 3 align-mode `MambaSpec` groups. Two hard blockers, one follow-up:

1. **Multi-group branch rejected the layout twice**: it only accepted raw `FullAttentionSpec`/`MambaSpec` group specs (the MLA group is `UniformTypeKVCacheSpecs`-wrapped, unlike the single-group branch which already unwraps it), and it required identical logical block sizes across *all* groups (8960 vs the kpool tail's 4).
2. **Registration mis-measured the kpool-compressed indexer K cache**: it stores one state per `tokens_per_state=4` tokens, so a 64-state kernel block spans 256 tokens and the physical/logical split ratio is 35, not 8960/64=140 — `physical block count must be divisible by physical/logical split ratio (17885, 140)`.

## Changes (`python/pegaflow/connector/` only — no Rust/proto changes)

- `common.py`: classify `KpoolTailSpec` groups (raw or uniformly wrapped) as **scratch groups**. They are excluded from hash-group election and the block-size uniformity check. The kpool tail is `prefix_cacheable=False`, holds only the in-progress pool, and every external hit boundary is pool-aligned (hash granularity 8960, `index_kpool=4`), so it holds no prefix state and is never registered/saved/loaded — the recompute after a load repopulates it. Also unwrap uniformly grouped MLA specs in the multi-group branch so the MLA group is recognized as the hash group.
- `worker.py`: skip scratch layers in `register_kv_caches`; `_infer_kv_cache_registration` takes per-layer `tokens_per_state` so compressed indexer caches compute the split ratio in token units.
- `scheduler.py`: scratch groups join recurrent groups in the non-positional category — excluded from the saveable-length clamp in `_consume_full_block_saves` (their 1-block mirror would otherwise clamp every save to ~1 block) and emitted as all-`None` destinations in `_load_block_ids_by_group` (their 1-entry block table would otherwise fail the load availability check).

## Tests

- New unit tests: `test_accepts_glm53_flash_layout_with_kpool_tail_scratch` (5-group GLM-5.3 layout: classification, hash group, storage ids) and `test_kpool_indexer_cache_uses_tokens_per_state` (17885 blocks / ratio 35 regression); existing SimpleNamespace cache-group doubles updated for the new field.
- Unit gate on the GPU pod (source-only, no torch/vLLM): `PYTHONPATH=. uv run --isolated --no-project --with pytest --with numpy --with requests pytest -q` → **306 passed, 13 deselected**. (Running the suite against the pod's *installed* vLLM fails because the real frozen-dataclass specs can't be mutated by these stub-style tests — pre-existing environmental issue, unrelated to this change.)
- `prek run` python hooks (ruff/isort/codespell/typos/whitespace) pass. Rust untouched; cargo hooks not run.

## E2E on GB300 pod (vLLM 0.29.1rc1 + GLM-5.3-Flash-NVFP4, TP1/EP1, fp8 KV, MTP spec-5, pegaflow-server 50GiB)

- Server healthy with connector; registration logs show the MLA + indexer + Mamba groups registered and the indexer layers at `physical_blocks_per_logical=35` (511 logical blocks).
- Cold/warm TTFT with an 11k-token prompt: **2067 ms → ~170 ms** (~12x); connector log `cache_lookup: hit_blocks=1 hit_tokens=8960`; server-side `cache_block_hits_total` and 214.6 MB loaded from the RAM tier; `pega_load_failure_total = pega_save_failure_total = 0`.
- **Restart proof that hits are external**: after killing vLLM (fresh process, empty GPU cache, new instance id), the *first* request for the same prompt hits (`hit_blocks=1`) — local HMA prefix lookup is disabled by the connector, so this can only come from the pegaflow server.
- Correctness: a secret planted inside the loaded block is recalled correctly after cross-restart loads; output matches a no-connector reference server. (Greedy continuations of the repetitive filler occasionally flip between two phrasings — reproduced identically on the no-connector server, i.e. engine numerics on partial-prefill resume, not cache corruption.)

Note: this PR was prepared with AI assistance (Kimi Code); the E2E bring-up and verification above were run on the GB300 dev pod as described.


Kevin-XiongC:
## Accuracy eval under load (evalscope, concurrency 32, temperature 0, max_tokens 16384)

Setup: GLM-5.3-Flash-NVFP4 TP1/EP1 on one GB300, full GSM8K (1319) + full AIME-2025 (30), served via OpenAI chat API.

- **pega server (this PR)** ran with concurrent interference on the same GPU: 512 random 9216-token requests (concurrency 12, one full pega block saved each, ~110 GB written into a 50 GiB pool so eviction was exercised) plus a repeated 14.4k-token prompt looped 323 times (a pega load hit per iteration). Local GPU prefix hits were 0.0% throughout (the connector disables local HMA prefix lookup), so every cross-request hit came from PegaFlow — external prefix cache hit rate peaked at 81.4%/53.3% in the loop windows.
- **baseline server**: identical flags minus the connector, no interference, on a second GB300.

| | GSM8K (1319) | AIME-2025 (30) |
|---|---|---|
| PegaKVConnector + interference | **97.3%** | **50.0%** |
| baseline (no connector) | 97.0% | 56.7% |
| PegaKVConnector, no interference | 97.1% | — |

- GSM8K: no accuracy loss (Δ +0.3pp, within greedy-decoding batching noise).
- AIME: 15/30 vs 17/30 — at n=30 this is within noise (±2 problems ≈ ±6.7pp); a `--repeats 4` avg@4 run would tighten this if needed.
- Zero errors on either side during the storm: `pega_load_failure_total = pega_save_failure_total = 0`, all 323 interference-loop requests succeeded, no eval request timed out or 5xx'd.

Kevin-XiongC:
**Correction to my accuracy-eval comment above** (the hit-traffic attribution was wrong):

Authoritative counters from the eval window show `vllm:external_prefix_cache_hits_total = 26,880` tokens at the time — exactly the 3 blocks (3 × 8960) from the earlier restart-proof demo, i.e. **zero pega load hits occurred during the interference + eval window itself**. The repeated-prompt loop actually missed every iteration, and the "external hit rate 81.4%/53.3%" log windows were the decaying cumulative average from the demo hits, not loop activity.

Why the loop missed: (1) save→queryable visibility latency is 5–10 s (verified by a controlled probe: identical 13k-token prompt at t=0/5s/10s → miss, miss, **hit**), while the loop polled every ~1.6 s; (2) the 512×9216-token random storm wrote ~2× the 50 GiB pool through it, so the loop's block was evicted between save and reuse. Local GPU prefix hits stayed 0.0% throughout, as designed.

So the accuracy numbers (GSM8K 97.3% vs 97.0% baseline; AIME 50.0% vs 56.7%) measured "connector online + heavy GPU contention + massive pega **save** churn", not "pega load hits under pressure". The TTFT hit-path evidence in the PR body (cold 2067 ms → hit ~170 ms, cross-restart hit) remains valid. A hit-under-load accuracy variant would need a bigger pool or slower unique-block churn plus a >10 s reuse delay; happy to run that as a follow-up.

Kevin-XiongC:
## Two-pass restart test: GSM8K accuracy served 100% from PegaFlow cache

Setup: GSM8K full (1319 problems), concurrency 32, temperature 0, `few_shot_num=64` so every prompt shares an identical ~14.5k-token prefix (> one 8960-token block — short default prompts never fill a block and would bypass the connector entirely).

- **Pass 1** (cold pega cache): 97.4% (1285/1319). Within the pass, once the first save became visible (~5–10 s save→queryable latency), every later request loaded the shared prefix block from pega: 11,818,240 external hit tokens = 1319 × 8960 exactly.
- **Kill vLLM, wipe GPU cache, restart** (fresh process, new instance id; local prefix lookup stays disabled by the connector, so local hits are impossible).
- **Pass 2** (same prompts): **every one of the 1319 requests loaded its 8960-token prefix block + mamba boundary checkpoints from the pega server** (`external_prefix_cache_hits_total = 11,818,240` = 1319 × 8960 again; 1174 successful loads, 0 failures, 0 saves needed since blocks were already stored).

| | GSM8K accuracy |
|---|---|
| Pass 1 (cold cache) | 97.4% |
| **Pass 2 (100% pega-fed prefix)** | **97.0%** |
| No-connector reference server | 97.0% |

Δ = 0.4pp (5/1319), within the run-to-run wobble we also observe between two no-connector runs (greedy decoding, batching/partial-prefill numerics). Conclusion: serving prefixes from PegaFlow — including mamba state restoration across a process restart — is accuracy-neutral at GSM8K scale.

GentleCold (approved):
LGTM

